PR #9, opened 2016-04-22
fix typos in lab w01

Zolomon:
Fixed all typos I could find in this file.

Should I add all recompiled PDF files to commits of this kind now and in the future?

### DIFF
--- a/compendium/modules/w01-lab.tex
+++ b/compendium/modules/w01-lab.tex
@@ -35,7 +35,7 @@ fram; vänster
 
 \Subtask Prova att ändra på ordningen mellan satserna och använd den \emph{gula} play-knappen  (programspårning) för att studera vad som händer. Klicka på satser i ditt program och på rutor i programspårningen och se vad som händer.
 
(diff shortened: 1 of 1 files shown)


bjornregnell:
Thanks! No, don't add pdfs: keep the PR clean with an as small change set as possible/reasonable as motivated the fixed file(s). I'll do the recompile after merge and push pdfs. But good that you verified that it compiles before creating a PR! You can check out your changes to your pdf before you synch your fork to avoid conflicts.

bjornregnell:
I added an explanation in relation to your Q in README under how to contribute; pls check to see if it now might help more.

Zolomon:
It looks quite clear, IMHO.

Thanks!

Zolomon:
It just occurred to me that maybe a comment similar to would be good to give concrete examples: 

"To see examples of acceptable PR:s, take a look at [previously accepted ones](https://github.com/lunduniversity/introprog/pulls?q=is%3Apr+is%3Aclosed)."

bjornregnell:
Good idea! PR on README.md welcome! ;)

PR #26, opened 2016-05-02
Update w09-lab.tex

lindskog:
First draft of labdescriptions for w09-lab maze.


bjornregnell:
This PR didn't compile. I've fixed the problems (coming in next commit, need to inser backslash in latex before _ and # ). Please check if it compiles before you commit next time.

PR #33, opened 2016-05-09
Added self to contributors, fixed spelling of license file

ErikBjare:


ErikBjare:
Nope, my bad.

Apparently licence is an acceptable (and in this case likely the preferred) spelling: http://grammarist.com/spelling/licence-license/

PR #37, opened 2016-05-12
Corrected a minor mistake in function call

Zorobay:
The function f was not correctly called. The variable y does not exist and f only takes 1 argument.


bjornregnell:
Thanks- good spotting this misstake!
Your PR had conflicts as you included pdf files. Next time only commit the tex file and discard changes on the .pdf (using discard changes in github desktop or using git checkout path/to/files/*.pdf)
Ok?

Zorobay:
Gotcha! 👍

PR #93, opened 2016-06-11
Rescaled images

MechaMagpie:
Rescaled images from 2500p width to 750p width


bjornregnell:
@MechaMagpie  Perfect! Many thnx!

PR #109, opened 2016-07-08
Removed .scala in README text

gustavcedersjo:


gustavcedersjo:
I haven't checked the commands, but I believe this will work.

PR #116, opened 2016-07-22
added w07-solutions and fixed w07-exercise

tpi14owi:

PR #117, opened 2016-07-23
Minor errors in lab w03 corrected, and started w10 solutions

Zorobay:


bjornregnell:
Your PR does not compile :(
I will try to fix it; but next time pls check first if it compiles before sending a PR!

Zorobay:
I was afraid of that, but the reason it does not compile is that the version I pulled did not compile :( Are you using sbt to check if it compiles? Maybe I should start using that...

bjornregnell:
The reason that your repo did not compile the first time was another broken PR #116, but I fixed that. You should have waited until your repo compiled before you sent your PR. 
I mostly use pdflatex biut sometimes sbt -- but you don't need sbt to check if it compiles -- just use pdflatex on compendium.tex. But sbt is ofcourse nice when you get around to user it, esp if you want to build the auto-generated files from the scala code in plan/ etc.

Zorobay:
Yeah okay, I'll be more careful next time! :+1:

PR #120, opened 2016-07-26
Fixed travis config

ErikBjare:
See working build here: https://travis-ci.org/ErikBjare/introprog/builds/147437600

The issue was not getting a recent `texlive-latex-extra` package, I switched to a full VM (with trusty/14.04) and installed it and now it runs fine.

Build time is ~10min, most of that time (~400s) is spent installing packages via APT. Maybe this can be reduced by not installing texlive-full and instead choosing the necessary packages with better precision. But 10min isn't that bad, so for now it'll do.


bjornregnell:
GREAT! :+1:

bjornregnell:
Build time is ok for now; I've no idea on how to, with limited human effort, check what exact latex packages that are needed, but let's investigate that in due course and see if we can cut down on the build time eventually.

ErikBjare:
It's nothing to worry about really. It works, that's what counts. Selecting
packages is a pain, suddenly you want another one etc. This way is better
because it's closer to the environment we compile in ourselves.

The next step is to upload the builds automatically for tags/releases so we
can get rid of the PDFs in the repo once and for all. I will get on it
after I've done something similar for my ActivityWatch project.

On Jul 27, 2016 13:45, "Bjorn Regnell" <notifications@github.com> wrote:

Build time is ok for now; I've no idea on how to, with limited human
effort, check what exact latex packages that are needed, but let's
investigate that in due course and see if we can cut down on the build time
eventually.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/lunduniversity/introprog/pull/120#issuecomment-235562036>,
or mute the thread

PR #126, opened 2016-08-01
Solutions to w09

hnrklssn:
First draft of the solutions to w09 basic and extra exercises
Solutions removed from the w09 exercise instructions
Compiled the new version of the compendium


bjornregnell:
Dice is actually singular:
http://www.oxforddictionaries.com/definition/english/dice
"throw the dice"

PR #148, opened 2016-08-18
Fixed some errors in w10, and added new headings to w10-solutions.tex

Zorobay:


bjornregnell:
Can you redo this PR but *without* compendium.pdf (you can safely discard changes from the pdf) as it causes conflicts with the pushed binary. (Never include pdfs in PR:s)
Rember to update your fork just before sending the PR.

PR #155, opened 2016-08-21
Added task headers to w10-sorting

Zorobay:


bjornregnell:
Thanks for PR! And very good hat all checks passed. Unfortunately there is a pdf among the commits included in this PR so I will refrain from merging and close it. Can you make a new one without the compendium.pdf ?

bjornregnell:
If you have trouble doing it this time I can take the time to cherry-pick the other files, if you resend it still with a compendium.pdf but it would be good if that was avoidable. Let me know how it goes.

PR #167, opened 2016-08-27

HSIHF:
I have tested the w08-lab and I have worked as a DEV for the w07-lab. Furthermore I have proofread several of the texts (labs & exercises) and fixed all of the typos/inconsistencies that I could find in these texts.


bjornregnell:
Many thanks for your contributions!!

PR #184, opened 2016-09-04

perssonsimon1:


bjornregnell:
Thanks Simon! Next time: the commit msg is for the records, so write something that is useful when looking back. Eg. "update list of contributors, added personsimon1". Ok?

PR #192, opened 2016-09-19
added some solutions to w04

Zorobay:


bjornregnell:
Better to wait for Viktor Claesson; he has created all solutions since before but these were for some strange reason never merged, but he is on it -- he just needs to update against upstream an provide a pull request...

PR #198, opened 2016-09-27
Updated if-statement to check if key exists to prevent NoSuchElementE…

CuddlareN:
…xception

If the key that we want to access, for example Royal flush, does not exist in the map scala will throw an exception and we must therefore check if it exists for the program to work


bjornregnell:
Thanks for contributing! I guess this PR was closed for a reason; it would be good if you could write a short explanation in a comment of why you closed it?

CuddlareN:
Accidentally made the PR to this repo instead of a personal repo where it was supposed to be made to

PR #204, opened 2016-10-05
Update w06-exercise.tex

dat12fda:
class Rymdvarelse anropades som RymdVarelse, detta fel var ej tänkt att finnas:
Rad 41, 50, 52


bjornregnell:
Thanks! Next time: write commit msg in english and explain in the heading what is fixed.

PR #206, opened 2016-10-07
Changed Array to fill with 4 zeros

AdamBarvesten:
val ys = Array[Int](n) Fills the Array with one 4

val ys = Array.fill[Int](n)(0) Fills the Array with four 0's which is wanted in this task.

### DIFF
--- a/compendium/modules/w05-solutions.tex
+++ b/compendium/modules/w05-solutions.tex
@@ -111,7 +111,7 @@ zs = Array(84, 2, 3)\\
 \Subtask \begin{Code}
 def seqReverseCopy(xs: Array[Int]): Array[Int] = {
   val n = xs.size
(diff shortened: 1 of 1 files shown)


bjornregnell:
Good catch! It is more idiomatic to remove the redundant type annotation in this case as the zero is obviously an Int, so in order not to confuse people that it is needed. 
`val ys = Array.fill[Int](n)(0)`
Please update and come with another PR. Thanks!

bjornregnell:
Sorry, forgot to state the idiomatic way to do it:
`val ys = Array.fill(n)(0)`

bjornregnell:
I'm reopening and then merging ; we can live with this for now; later add comment that [Int] is not strictly needed.

PR #207, opened 2016-10-07
Once again changed Array.fill with 4 elements(0)

AdamBarvesten:


bjornregnell:
reopening and merging; later we can add that [Int] is not strictly needed

PR #209, opened 2016-10-09
fixed #205 by updating Turtle.scala in the workspace

### LINKED ISSUE #205: w06-turtlegraphics, left out return type
In the definition for Turtle in turtlegraphics all methods have their return type explicitly defined except for the jumpTo method. Is there a reason for this or is it just a inconsistency that should be corrected?

emilwihlander:


bjornregnell:
sorry have to close this as I have taken the Turtle to the next level.. Thnx anyway!!

PR #214, opened 2016-11-13
fix maze path being dependent on CWD; use resource dir

vlthr:
I could not run the maze lab on {linux, windows} + IntelliJ (with sbt plugin) due to the use of `new File(".")` which in IntelliJ yields the root project (workspace/) rather than the submodule (w09_maze).

Normally, we could get around this by not explicitly relying on a directory path, but just using `getResource(filename)` to access the maze files. However, since the lab is explicitly teaching file handling rather than resource use, I instead fixed the problem by dynamically getting the directory where the resources are stored.

Tested and working on following platforms:
linux + sbt
linux + IntelliJ IDEA (with sbt plugin)
windows + IntelliJ IDEA (with sbt plugin)

What do you think? Is it confusing to the students to use a path which is somewhere in the target/ directory?

### DIFF
--- a/workspace/w09_maze/src/main/scala/maze/AMazeIngRace.scala
+++ b/workspace/w09_maze/src/main/scala/maze/AMazeIngRace.scala
@@ -3,6 +3,7 @@ package maze
 import cslib.window.SimpleWindow
 import java.awt.Color
 import java.io.File
+import java.nio.file.Paths
 
 object AMazeIngRace {
   def printMazesFromDir(filepath: String): Unit = {
@@ -54,8 +55,10 @@ object AMazeIngRace {
     }
   }
 
+  def getMazeDirectoryPath : String = Paths.get(getClass().getResource("/maze1.txt").toURI).toFile.getParent+File.separator
+
   def main(args: Array[String]): Unit = {
-    val path = (new File(".")).getCanonicalPath + "/src/main/resources/"
+    val path = getMazeDirectoryPath
     val w = new SimpleWindow(800, 800, "A-Maze-Ing Race")
     printMazesFromDir(path)
     w.moveTo(10, 100)


bjornregnell:
Would it be possible to get the path without explicitly relying on a specific file name?
Something like  getClass.getResource("/") or similar in a def getResourceDirPath which perhaps is a better name than getMazeDirectoryPath?

bjornregnell:
Perhaps like this and we don't have to rely on java.nio and make it even more minimal/cleaner:
def slash = java.io.File.separator
def resourcePath = getClass.getResource(slash).getPath

I'll merge your PR and then play with it.

bjornregnell:
I just tried it on Windows and backslash in getClass.getResource(slash)  gives null in the REPL...
So Windows should not use its own file separator -- this is actually quit funny ;)
This seems to work on Windows and Linux in the REPL and gives the current dir:
def resourcePath = getClass.getResource("/").getPath

vlthr:
I just tried your most recent version and can confirm that it works in both linux and windows under IntelliJ and SBT. Much cleaner!

bjornregnell:
Thnx for testing! A releif to know that it works :)
Still puzzled why the first solution did work on my IntelliJ-box though...
Could be some setting in IntelliJ; perhaps you have changed how resource folders are handled in your settings? Is it perhaps a Maven/sbt/gradle dir structure clash?

vlthr:
I was wondering about that too! For both linux and windows I was using a freshly installed IntelliJ version on a freshly cloned repo.

The only possible culprits that come to my mind are:
- Are we using the same project root? I imported the entire workspace/ as one project, with each lab being a submodule. The path that was being tried earlier was `..../workspace/src/main/resources/maze1.txt`, which is correct except for the missing `w09_maze`. This means that `new File(".")` was giving me the path to `workspace/`. If I had imported w09_maze directly I imagine that would have given me the correct path by default.
- Are you also using the SBT plugin for IntelliJ to autoimport settings from build.sbt? I haven't tried without it, but I imagine that it might change my run configurations.

bjornregnell:
Aha! The first explanation holds: I was importing this project directly in an IntelliJ instance not the entire workspace... Very good that you spotted this and the current solution should be robust to all of that.
(I was not using the sbt plugin this time, but probably sbt would make things easier, but as students seems often not using sbt, I don't when testing...)

vlthr:
Ahh, that makes sense! The SBT plugin is nice, but I understand that many of the students find it confusing to layer two build managers (IntelliJ + SBT) on top of eachother when they're not entirely sure how either one works. I should set up a config more like what the students are using to test on.

Issue #217, opened 2016-11-18
w09-maze: Can't load mazes when path contains spaces


jos-rosenqvist-3812:
In the `maze` lab, loading the maze files will fail with a `FileNotFoundException` if the path to the files contains spaces.

`getResourcesPath` returns a string that contains `%20` instead of spaces. Doing `.replace("%20", " ")` on the string makes maze loading work correctly, but it doesn't feel like a proper solution. Maybe there are other characters that also get returned as `%` escape sequences.

bjornregnell:
Good catch! Yes this was a quick hack and not general. If you want you could investigate these SO answers (probably getAbsolutePath() on fixes the issue?):
http://stackoverflow.com/questions/6164448/convert-url-to-normal-windows-filename-java 
and then perhaps come with a Pull Request on the relevant workspace files of this repo to fix this issue?

PR #218, opened 2016-11-18
Make maze file paths work with special characters

jos-rosenqvist-3812:
The first commit fixes issue #217.

The second commit refactors the path handling so that the first commit's extra `"/"` in `getResourcePath` isn't needed. If this commit isn't wanted, let me know, then I'll remove it from the PR.


bjornregnell:
Thanx! Would be nice with a PR on contributors.tex with your name added in the list... :+1:

PR #219, opened 2016-11-18
Add myself to contributors

jos-rosenqvist-3812:
See PR #218.


bjornregnell:
Thanx! If you want to contribute more, you can take a look at the new survey2 (a simplified version of the old survey with less command parsing included) and check the file management there. The same solution as you did in maze might not be entirely applicable in survey2 as the user inputs file names from a command prompt... The "current dir" should work on all those various platforms Win/Mac/Linux Eclipse/IntelliJ/Sbt and in straight terminal with scalac compile... Maybe too much to be able to cover them all; then prioritize Win/Mac/Linux Eclipse/IntelliJ ...  I think its ok if it only works with file names without spaces, but you might want to give it a go?

PR #220, opened 2016-11-18
survey2: Improve path handling

jos-rosenqvist-3812:
- The code for getting the resource directory is now only in one place instead of three.
- The same fix as in 45ec99f is applied, making special characters work in paths. (Writing paths with spaces during the command loop is impossible for command parsing reasons, but spaces do work in paths provided as arguments to main.)
- Instead of having hardcoded checks for strings starting with "http", the program now determines if something is a URL by trying to create a URL object from it. If there is no valid protocol, a MalformedURLException is thrown, which we catch.
- Path.resolve is now used instead of string concatenation, which lets the user optionally enter absolute paths.

I don't have access to many testing environments here, but it seems to work fine in at least Windows+IntelliJ.

Is it intentional that the `save` method doesn't use the resource directory as a base for relative paths, unlike `load`?

### DIFF
--- a/workspace/w10_survey2/src/main/scala/stats/Command.scala
+++ b/workspace/w10_survey2/src/main/scala/stats/Command.scala
@@ -39,10 +39,8 @@ object Command {
       |             Terminates the program""".stripMargin
 
   def load(uri: String, sep: String): Table = {
-    lazy val path = getClass.getResource("/").getPath
-    val location = if (uri.startsWith("http")) uri else s"$path$uri"
-    println(s"""Loading $location with separator '$sep' to table ...\n""")
-    val t =  Table.fromFile(location, sep)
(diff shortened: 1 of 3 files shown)


bjornregnell:
Nice! 
> Is it intentional that the save method doesn't use the resource directory
No reason save is so simple other than that it works and it is simple :) 
-- it uses whatever current dir is the current dir and that's fine if you e.g. compile and run in terminal. 
Do you think there is a big downside to this?

jos-rosenqvist-3812:
No, I don't think there's any significant downside to it. Unlike load, it doesn't matter exactly where the saved file ends up, as long as people are able to find it.

bjornregnell:
Ok, we'll leave it as is then.

PR #223, opened 2016-11-23
Fix compilation errors in search/sort chapter exercises

### LINKED ISSUE #222: Search/sort task 15 b) doesn't compile
In the search/sort chapter, task 15 b) is written as if the provided `isSorted` function works correctly, but it fails compilation with `error: No implicit Ordering defined for T.`

Task 15 d) contains the exact same code, but correctly says that there is a problem with the code.

jos-rosenqvist-3812:
Fixes #222 and an issue that doesn't have an associated issue report.

### DIFF
--- a/compendium/modules/w11-exercise.tex
+++ b/compendium/modules/w11-exercise.tex
@@ -99,7 +99,7 @@ def timed[T](code: => T): (T, Long) = {
   val now = System.nanoTime
   val result = code
   val elapsed = System.nanoTime - now
-  println(s"\ntime: ${elapsed / 1e6}$ ms") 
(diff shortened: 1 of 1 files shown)


bjornregnell:
Thanks for PR! I was working on improving the exercise. Would be very good if you could test the latest version and let me know if it makes more sense? Commit here: https://github.com/lunduniversity/introprog/commit/a20dd8e2ef64edf7baa4073acbff9e97622ea362
and in latest pushed compendium.pdf

jos-rosenqvist-3812:
Looks nice. It makes sense that it first shows the non-working version and then how to fix it (instead of "going backwards" like the old version did, so to speak), and it also explains what's going on more clearly.

PR #233, opened 2016-12-06
Fix inconsistencies in bank

vlthr:
As pointed out by the students, there were some inconsistencies in the compendium about which classes were required. 

Main changes:
* All mentions of Transaction have now been updated to refer to BankEvent (which was originally renamed since some of the events are not strictly transactions).
* Fixed mentions of singleton objects that were being referred to as classes.


bjornregnell:
Thanx!!

PR #239, opened 2016-12-09
Fix appendix typos and other minor problems

jos-rosenqvist-3812:


bjornregnell:
Many thanks for all these commits! I'm wondering what is best wrt spacing:
http://tex.stackexchange.com/questions/116528/vs-after-abbreviation-read-backslash-at-vs-backslash-space
Perhaps the \sbt macro should be redefined instead to \newcommand{\sbt}{\texttt{sbt}\@} if I understood the above Q&A correctly.

jos-rosenqvist-3812:
I'll try that out after the weekend.

If I'm going to edit the PR, do you want me to edit the old commit and force push, or just add a new commit? (I'm fine with either.)

bjornregnell:
Thanks! Just add a new commit; I'll merge all your existing PR:s in consecutive order during the weekend.

jos-rosenqvist-3812:
The `\@` is used for forcing inter-sentence spacing after a period. We don't have a period here, so there doesn't seem to be any point to using `\@`. (Does Swedish even use different spacing inside sentences than between sentences?)

bjornregnell:
Ok, good; then we skip \@; but it would be nice to define \sbt so that \sbt\ is not needed if possible. Or we just keep it as is now.

PR #251, opened 2016-12-15
Fixed #250 and fixed inconsistencies in w12_Life

### LINKED ISSUE #250: w12_Life doesn't import correctly
It doesn't import correctly since the .classpath file doesn't contain a link to "/src/main/scala/". The files aren't located in "/src/main/scala/" but directly in "/src/".

emilwihlander:
Fixed #250.
moved source files to "/src/main/scala/"
added missing curly bracket in "CyclicRule.scala"
fixed ".classpath"
fixed inconsistencies.


bjornregnell:
@ErikBjare Could you, as dev lead on this, have a quick look at the changed files of this PR which addresses the issue #250 and report a LGTM here if it looks good to you?

emilwihlander:
@ErikBjare was with me while I fixed the problem and also did one of the commits as you see

bjornregnell:
Aha, I missed that! Then I'll just merget it. Many thanks!

PR #257, opened 2017-04-18
Replace W3Schools links

jos-rosenqvist-3812:
W3Schools is not often regarded as a good resource, so I'm replacing the links to it with a link to the more respected MDN (Mozilla Development Network).

### DIFF
--- a/compendium/modules/w13-exercise.tex
+++ b/compendium/modules/w13-exercise.tex
@@ -423,11 +423,7 @@ Hur stora tal klarar din server nu? Vad händer med servern när minnet tar slut
 
 \Task Fördjupa dig inom webbteknologi. 
     
-\Subtask Lär dig om HTML här: \url{http://www.w3schools.com/html/}
-
-\Subtask Lär dig om Javascript här: \url{http://www.w3schools.com/js/}
-
-\Subtask Lär dig om CSS här: \url{http://www.w3schools.com/css/}
+\Subtask Lär dig om HTML, CSS och JavaScript här: \url{https://developer.mozilla.org/en-US/docs/Learn}
 
 \Subtask Lär dig om Scala.JS här: \url{http://www.scala-js.org/}
 


ErikBjare:
I think it's great that you removed W3Schools, but is MDN really useful for beginners?

bjornregnell:
If W3School is not ok, it should be removed, but I'm curious to learn more about why W3School is "bad". Any references/links about it's "badness" to check-out?

ErikBjare:
@bjornregnell A lot of discussion about this has happened on HN over the years, [a Google search](https://www.google.se/search?q=w3school+hacker+news) gives you:

 - https://news.ycombinator.com/item?id=6422998
 - https://news.ycombinator.com/item?id=7230265
 - https://news.ycombinator.com/item?id=2104271

After reading some of the discussion, I concede that W3Schools might be one of the most newcomer friendly resources out there despite its flaws.

It might be best to keep W3Schools after all and simply add MDN to the list, at least until we find a good alternative that covers the needs of beginners.

jos-rosenqvist-3812:
@ErikBjare It seems like MDN is split into distinct parts, including "Web Technologies" and "Learn". The Web Technologies part is probably the one that's the most familiar to developers, but what this PR adds a link to is Learn, and that part of MDN is intended for people who are just starting out.

@bjornregnell Seems like Erik was a bit faster than me at responding to you - I was looking up some links in this very moment :)

bjornregnell:
@jos-rosenqvist-3812  I think it is best for now to add MDN but keep W3school, if you are OK with that?

jos-rosenqvist-3812:
That sounds okay. I can make that change to this PR after the weekend.

Is it fine if I at the same time replace the three W3Schools links with just one, so that the number of links in the section becomes more manageable?

bjornregnell:
Sure, sounds like a good idea!

bjornregnell:
BTW, I'll soon make a big restructuring of the compendium in order to separate it into two parts. While I do that it's probably best to wait with any other changes... If I start that work in the wekend (probaly not) then I'll accept this PR as is and you can fix it later. I'll let you know if I start some big refactorings to let you hold your gun...

bjornregnell:
merging this; I'm about to split the compendium; I'll re-construct links to w3school

PR #260, opened 2017-05-04
Added "Build URL" for compendium

sedrakk:
Hi everyone,

I noticed that the link to PDF version of compendium is not working and I updated it.

Basically, the link I added is something we call "Build URL" - each time it's visited, [latexonline.cc](https://latexonline.cc) will fetch all TeX files from the repository (if they were updated) and create the PDF. You won’t need to re-generate PDF each time its sources are updated. And if there are no updates, the service will return cached version.

We recently updated latexonline.cc to be more convenient for the users and I thought your repository is a good example where the service will be useful.

Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ This is work in progress; the first instance of the course starts this fall seme
 
 ## How to use this repo
 
-* Download the latest version of the [compendium](https://github.com/lunduniversity/introprog/raw/master/compendium/compendium.pdf).
+* Download the latest version of the compendium via [LaTeX.Online](https://latexonline.cc/compile?git=https://github.com/lunduniversity/introprog&target=compendium/compendium.tex&command=pdflatex).
 
 * Download the [workspace](https://github.com/lunduniversity/introprog/blob/master/lib/workspace.zip) to be used with your favorite code editor and IDE. See instructions in appendices in the compendium.
 


ErikBjare:
That's pretty neat, and thanks for disclosing your affiliation!

We're actually using Travis for continuous integration so we're generating the PDF every commit anyway and plan to have it automatically uploaded somewhere on every successful build. This might be a good solution in the meantime however. What do you think @bjornregnell?

bjornregnell:
Cool! Works like a charm!

I wonder: it seems as if it re-compiles even if the files in the repo hasn't changed, so the pdf is not cached if I understand it correctly? 

As the repo has grown big due to too many binaries changing on commits, I have just recently moved the changing pdf:s to the release section (after rewriting the repo history to kill the binaries):
https://github.com/lunduniversity/introprog/releases

I will also scp them via a script to our university web fileadmin server and will cache them there as well for redundancy. But for snapshots latexonline seems like a very good solution!

And, many thanx @sedrakk for spotting the broken link to the pdf in the README!

BTW: A progress bar or something indicating that the compilation is progressing would be cool!

aslushnikov:
> I wonder: it seems as if it re-compiles even if the files in the repo hasn't changed, so the pdf is not cached if I understand it correctly?

@bjornregnell we do cache pdf, relying on git SHA of the master branch. So if you change anything in your git history, we'll recompile. We also purge cache if the pdf is not accessed during 24 hours just to save on space. 

If it doesn't work like this - it's a bug! Please, let us know :)

> BTW: A progress bar or something indicating that the compilation is progressing would be cool!

Good call, filed this as https://github.com/aslushnikov/latex-online/issues/29

bjornregnell:
> If it doesn't work like this - it's a bug
No, it really seems to work, but I was confused by the message on the web page that said "compiling...".
I think that if there is already a pfd cached then the web page should say something like "Opening cached pdf..." so that the user does not think that the compilation is on-going when it's not.

PR #279, opened 2017-08-25
Fix outdated information

jos-rosenqvist-3812:
This PR focuses on fixing outdated information that will be in the printed LP1 compendium and the LP1 part of the screen-optimized compendium. There are still some outdated details in slides that aren't included in any version of the compendium.

As a side note, are we going to have a new VM image for this year? I didn't touch the compendium's list of program versions that are included in the VM image, because I wasn't sure if the image had been updated or not.


slides/body/lect-w07-classes.tex
```diff
@@ -614,7 +613,6 @@
 \item en automatisk \Emph{\texttt{toString}} som visar klassparametrarnas värde, 
 \item ett automatiskt \Emph{kompanjonsobjekt} med \Emph{fabriksmetod} så du slipper skriva \code{new},
 \item automatiska metoden \Emph{\texttt{copy}} för att skapa kopior med andra attributvärden, m.m...
-\item[] (Mer om detta i w06 \& w11, men är du nyfiken, kolla på uppgift 2d) på sid 261.)
```
jos-rosenqvist-3812: Keeping the w06 reference doesn't make much sense, but should we keep the reference to the future chapter+exercise? If so, how should we handle the page number now that the printed compendium doesn't have the same page numbers as the digital compendium?

bjornregnell:
All lectures past w01 may contain inconsistencies; they will most likely not be fixed until the week before each lecture....

PR #281, opened 2017-08-25
Reword w06 footnote about Swedish sorting

jos-rosenqvist-3812:
That Swedish letters don't sort as expected is not just because of historical reasons – it is impossible to have one sorting order that's correct for all languages, so even if we were to make a new order, it'll be wrong for some amount of languages no matter what. As an example, Ö is the last letter of the alphabet in Swedish, but German places it earlier in the alphabet, so making the order correct for Swedish would make it wrong for German.

(Of course, it's possible to write programs that select different sorting orders depending on e.g. a language setting, but this part of the compendium is about the "built-in" order that Char has.)

### DIFF
--- a/compendium/modules/w06-sequences-exercise.tex
+++ b/compendium/modules/w06-sequences-exercise.tex
@@ -585,7 +585,7 @@ case class Person(förnamn: String, efternamn: String)
 val ps = Vector(Person("Kim","Ung"), Person("kamrat", "Clementin"))
 \end{Code}
 Deklarera ovan i REPL och para ihop uttryck nedan med rätt resultat.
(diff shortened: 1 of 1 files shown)


bjornregnell:
Well the footnot doesn't say that there exists a non-strange ordering from a Swedish point of view... Actually the locale setting of the JVM can control sort ordering. I'd like to keep the ref to the Swedish wikipedia page on ASCII. Is there a footnote phrasing that you could live with?

jos-rosenqvist-3812:
What do you think about something like this? "Ordningen kommer ursprungligen från föråldrade teckenkodningsstandarder: https://sv.wikipedia.org/wiki/ASCII"

bjornregnell:
That's perfect. Do you want to amend this PR or do you prefer that I accept this and you make the change later?

jos-rosenqvist-3812:
I've amended the PR.

PR #282, opened 2017-08-25
Use hash of quiz name as seed for shuffling

jos-rosenqvist-3812:
This prevents all the existing quizzes from being re-shuffled every time the quizzes are rebuilt, which should make the future commit history cleaner.

By default, the seed used for shuffling is a hash of the name of the quiz, but it's also possible to specify a seed manually.

### DIFF
--- a/quiz/QuizData.scala
+++ b/quiz/QuizData.scala
@@ -2,14 +2,14 @@ package genquiz  // see types in package object in TypesAndUtils.scala
 
 object QuizData {  // to generate tables for a concept connection quizes in latex
 
-  def concepts(q: QuizName): Concepts =
+  def concepts(q: QuizID): Concepts =
     quizes(q).zipWithIndex.map { case ((k,v), i) => ((k.trim,i), (v.trim,i)) }
 
-  def quizNames: Vector[QuizName] = quizes.keys.toVector
+  def quizIDs: Vector[QuizID] = quizes.keys.toVector
 
-  val quizes = Map[QuizName, Vector[Concept]](
+  val quizes = Map[QuizID, Vector[Concept]](
 
-    "quiz-w01-concepts" -> Vector( //expressions
+    QuizID("quiz-w01-concepts") -> Vector( //expressions
       "litteral       " -> "anger ett specifikt datavärde",
       "sträng         " -> "en sekvens av tecken",
       "sats           " -> "en kodrad som gör något; kan särskiljas med semikolon",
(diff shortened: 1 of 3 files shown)


quiz/QuizData.scala
```diff
@@ -413,6 +413,6 @@ object QuizData {  // to generate tables for a concept connection quizes in late
       "" -> ""
     ).filter(_._1.trim.nonEmpty),
 
-    "" -> Vector()
-  ).filter(_._1.trim.nonEmpty)
+    QuizID("") -> Vector()
+  ).filter(_._1.name.trim.nonEmpty)
```
jos-rosenqvist-3812: Just curious: Why does each map/vector contain an empty entry that gets filtered out?

bjornregnell: Reason: I'm longing for the scala version that will include trailing commas:
http://docs.scala-lang.org/sips/completed/trailing-commas.html
But with my trailing empty hack I can copy paste stuff without worring about trailing commas...

quiz/TypesAndUtils.scala
```diff
@@ -1,6 +1,14 @@
 package object genquiz {
 
-  type QuizName = String   // to understand these types see data in QuizData.scala
+  // to understand the types below, see data in QuizData.scala
+
+  case class QuizID(name: String, seed: Int) {
+    def this(name: String) = this(name, name.hashCode)
```
bjornregnell: Why an auxiliary constructor that overlaps with apply-factory?

bjornregnell:
Thanx for cool stuff. I've not been able to review everything yet, but have an initial question on the use of an auxiliary constructor instead of apply in companion -- see review comment.

quiz/TypesAndUtils.scala
jos-rosenqvist-3812: At first, I only had this auxiliary constructor, but then I added the apply-factory once I noticed that this auxiliary constructor didn't work for the "usual" way of creating instances of a case class (i.e. using `apply` instead of `new`). Is it preferable to just have the apply-factory? If so, I don't have anything against with removing this auxiliary constructor.

jos-rosenqvist-3812: I've removed the auxiliary constructor.

bjornregnell:
We could have solved it using default arguments as well; I'll accept this and might change it later.

bjornregnell:
After some more though I realize that default argument would have required two parameter lists:
`case class QuizID(name: String)(val seed: Int = name.hashCode)` but then we need `QuizId("x")()` and that is not so good, so alternative apply is best and I'll leave it like so. Many thanks for your nice work!!

PR #286, opened 2017-08-30
Fix chapter 1 task 6 (chessboard grains)

jos-rosenqvist-3812:
The way that the compendium is worded right now, it's asking for the amount of grains in the 64th square, but both the answer (2^64 - 1) and the Wikipedia article are about the total amount of grains.

@bjornregnell This is the problem that I mentioned very briefly earlier today.


bjornregnell:
Ah! Good catch!

PR #287, opened 2017-08-31
Course Instructions - Fixing link

davidsoederberg:
Fixing the link for the study-group contract example, so it works correctly.


bjornregnell:
Nice catch!

PR #291, opened 2017-09-17
Update of W03 custom control structure example

kevinfurmanski:
A for-loop seems more reasonable since we know the number of iterations beforehand


bjornregnell:
Well, while is more primitive and thereby often the fastest you can get ;) But I see you point. Let me just check the context when I get the time so that there are no other reasons for shooding while...

kevinfurmanski:
There is probably no significant difference between the two, but the last control question for lab 1 does imply that you should use a for-loop if the number of iterations is known so for consistency it seems better to use the for-loop

bjornregnell:
Ok. Merging. Thanx!

PR #304, opened 2017-10-12
bank: Remove State class and remove confusing usage of the word "state" (plus misc rewording)

jos-rosenqvist-3812:
This will hopefully make the project easier to understand. See the commit messages for details.


bjornregnell:
There are merge conflicts preventing merge. Can you fix that? Then I'll merge.

jos-rosenqvist-3812:
The instructions don't point students towards the old design at all, and I don't expect them to use that old design, it's just that the old design isn't explicitly forbidden. I think most people will reach the conclusion of using a Vector (or something similar, like an ArrayBuffer) based on the instructions that are in this PR, and the instructions are made with that in mind.

bjornregnell:
OK great!

PR #308, opened 2017-10-14
Make maze file paths work with special characters

jos-rosenqvist-3812:
This PR is essentially the same as PR #218. It seems like that PR got undone when the maze lab was rewritten recently.


bjornregnell:
Ooops... That was a good catch on an overlooked regression. Sorry!

PR #313, opened 2017-10-16
bank: Various fixes

jos-rosenqvist-3812:


jos-rosenqvist-3812:
Yes, I think we're better off skipping it, since it doesn't really add anything to the task other than distraction.

PR #314, opened 2017-10-16
bank: Improve the progression of history-related tasks

jos-rosenqvist-3812:

PR #315, opened 2017-10-17
bank: Use scala.util.Try instead of try

jos-rosenqvist-3812:
As suggested in https://github.com/lunduniversity/introprog/pull/314#pullrequestreview-69519421

### DIFF
--- a/workspace/w13_bank_proj/src/main/scala/bank/BankEvent.scala
+++ b/workspace/w13_bank_proj/src/main/scala/bank/BankEvent.scala
@@ -1,5 +1,7 @@
 package bank
 
+import scala.util.Try
+
 sealed trait BankEvent {
   /**
    * Returns a string suitable for writing to log files.
@@ -42,7 +44,7 @@ object BankEvent {
    * Converts a string obtained from toLogFormat into a BankEvent object.
    */
   def fromLogFormat(str: String): BankEvent = {
-    try {
+    Try{
       val xs = str.split(' ')
       xs(0) match {
         case "D" => Deposit(xs(1).toInt, BigInt(xs(2)))
@@ -52,9 +54,9 @@ object BankEvent {
         case "E" => DeleteAccount(xs(1).toInt)
         case s => throw new IllegalArgumentException(s"Unknown BankEvent type: $str")
       }
-    } catch {
+    }.recover{
       case e @ (_: IndexOutOfBoundsException | _: NumberFormatException) =>
         throw new IllegalArgumentException(s"Invalid BankEvent string: $str", e)
-    }
+    }.get
   }
 }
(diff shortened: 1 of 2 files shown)


bjornregnell:
BTW: Perhaps BankEvent should include an ErrorEvent and capture an error message to include such things as the Try is catching in the history? I think a real bank would not just crash but make a record of exceptional stuff... So with this solution make the recoverblock return a new ErrorEvent which is a BankEvent or similar name. If you like the idea?

bjornregnell:
Se mina kommentarer. Vi diskuterar mitt förslag en runda till där innan jag mergar?
/B

On 2017-10-17 12:22, Jos Rosenqvist wrote:

As suggested in #314 (review)<https://github.com/lunduniversity/introprog/pull/314#pullrequestreview-69519421>

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/lunduniversity/introprog/pull/315

Commit Summary

  *   bank: Use scala.util.Try instead of try

File Changes

  *   M workspace/w13_bank_proj/src/main/scala/bank/BankEvent.scala<https://github.com/lunduniversity/introprog/pull/315/files#diff-0> (8)
  *   M workspace/w13_bank_proj/src/main/scala/bank/time/Date.scala<https://github.com/lunduniversity/introprog/pull/315/files#diff-1> (8)

Patch Links:

  *   https://github.com/lunduniversity/introprog/pull/315.patch
  *   https://github.com/lunduniversity/introprog/pull/315.diff

jos-rosenqvist-3812:
> Maybe the case-clause in recover is a bit cryptic to beginners using type annotations with OR and place holder syntax etc. I'd suggest case e: Exception => instead (if I'm not missing some tricky corner issue here..)

I agree, the syntax is hard to understand. `case e: Exception` should give us to identical behavior, but I think it's nice to write out the specific types of exceptions so that the intent is clear. Do you think it would be fine to make one case for each exception type (with duplicated code after the `=>`)? If not, we can go with `case e: Exception`.

> BTW: Perhaps BankEvent should include an ErrorEvent and capture an error message to include such things as the Try is catching in the history? I think a real bank would not just crash but make a record of exceptional stuff... So with this solution make the recoverblock return a new ErrorEvent which is a BankEvent or similar name. If you like the idea?

I think that's a bad idea. These functions are only used when loading history from a file at startup, so if an error occurs, it means that there's a formatting error in the file (or some bug in the parsing code). If that happens, we want to immediately make it clear to the user that something is wrong, so that they won't have to discover unexplained errors later such as accounts being missing even though they used to exist. Of course, a real bank should show some kind of proper error message rather than just crashing, but I don't think that's an especially useful addition to the project, since the main method won't really have much choice other than quitting after displaying the error message.

bjornregnell:
* OK. I see your points, skip the ErrorEvent idea. 
* One case for each exception is ok and then finally a case e: Exception => to catch the rest (but not too much...) if you like it like that?

bjornregnell:
I'll merge for now and you can make another PR with easier to read exception match.

PR #322, opened 2017-10-26
Correct facts about MIDI synthesizers

jos-rosenqvist-3812:
The majority of modern computers do not have sound cards that include hardware synthesizers. That was only really popular during the 90s. For the reference, Microsoft added a software synthesizer in Windows 98 and removed support for hardware synthesizers in Windows Vista.


bjornregnell:
You are perfectly right as usual and facts are facts :) although the original text did not explicitly say that the the synth in your computer is a *hardware* synths, infact many "hardware" synths today that are sold as physical boxes actually include software-implemented simulations so the distinction is kind of blurred nowadays I guess ;)

PR #336, opened 2017-11-15
Make interval calculation not depend on note range

ErikPrantare:


bjornregnell:
Thanks for you contribution @ErikPrantare ! Can you elaborate a bit on this as I from a quick glance don't really get the connection with the commit msg and your code, and you have not provided any further description. Can you explain a bit more which problem your are solving and how it works?

ErikPrantare:
Finding `mod(a, b)` for positive b can be done as `(a%b + b)%b`. The way
it's done here(`(n*b + a)%b` with large `n`) works in this specific case,
but it could create problems for example if `a < n*b`. `(a%b + b)%b` works
better in the general case, and I also find it more intuitive as it gets
rid of the magic number 12*11.

ludvigparsson:
what @ErikPrantare said is correct, except for the "but it could create problems for example if `a < n*b`" part. What I think he meant here is that it will be a problem if: a is negative and abs(a) > n*b

With this change the elements in the Vector that relativePitchClasses returns will never be negative, even if the Pitch intervall becomes greater than 12*11

ErikPrantare:
You are correct, I meant to write `n*b + a < 0`.

bjornregnell:
OK now I get it! Thanks!

PR #337, opened 2017-11-22
Sista raden av hangman skrivs inte ut

ErikKullberg:


bjornregnell:
@ErikKullberg  Have you tested this now in all edge cases so that +-1 errors etc. are avoided?

ErikKullberg:
I have tested it to the best of my abilities.

PR #339, opened 2017-12-01
Changed to correct order in text for W-A-S-D

rvh97:
Det var i fel ordning i förhållande till tangenterna

vänster-ner-höger-upp
blir:
upp-vänster-ner-höger


jos-rosenqvist-3812:
Thanks, I missed this when changing from ASDW to the more common WASD. Could you also fix the same problem with IJKL one paragraph down?

rvh97:
Yes, I fixed the order for the I-J-K-L text too

bjornregnell:
@rvh97 Thanks for spotting this! 
@jos-rosenqvist-3812 Thaks for reviewing and approving!

Issue #338, opened 2017-11-24
fix static Hangman.java and JShowInt.java


VictorWinberg:
Why is these two java classes: [Hangman.java](https://github.com/lunduniversity/introprog/blob/master/compendium/examples/scalajava/Hangman.java) and [JShowInt.java](https://github.com/lunduniversity/introprog/blob/master/compendium/examples/scalajava/JShowInt.java) like singleton static-only objects?

I think it is more correct instantiating the classes in the `main` method rather than making it singleton, i.e., in `main` writing `new Hangman().play`. However in the case for Hangman the download method could still be static, i.e., `new Hangman().play(Hangman.download(...))`.

Just as the case for `Scanner` you will instantiate the scanner by `new Scanner(System.in)` I think that they should learn instantiating their own classes.

bjornregnell:
There are other classes without static later so I think this a minor problem. Also in the javatext lab the ganme is required to be restarted so they need to take this step there. So i'm inclined to keep this as is.

PR #341, opened 2017-12-01
fix scalajava example hangman static to non-static

VictorWinberg:
Example on how [Hangman.java](https://github.com/lunduniversity/introprog/blob/master/compendium/examples/scalajava/Hangman.java) could be changed to fix issues like #338.

However it still exist several more, in my opinion, weird java singleton static-only classes that could be changed like this one.

What do you think @bjornregnell?

### DIFF
--- a/compendium/examples/scalajava/Hangman.java
+++ b/compendium/examples/scalajava/Hangman.java
@@ -5,7 +5,7 @@ import java.util.HashSet;
 import java.util.Scanner;
 
 public class Hangman {
-    private static String[] hangman = new String[]{
+    private String[] hangman = new String[]{
         " ======  ",
         " |/   |  ",
         " |    O  ",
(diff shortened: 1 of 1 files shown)


compendium/examples/scalajava/Hangman.java
```diff
@@ -103,13 +103,14 @@ public static void play(String secret){
     }
 
     public static void main(String[] args){
+        Hangman hangman = new Hangman();
```
VictorWinberg: This is the difference needed in `main`-methods when not using static methods by instantiating the class.

bjornregnell:
Thanks @VictorWinberg for this contribution! Yes, this exercise was not meant to be Java-idiomatic but illustrate the semantic equivalence between singletons in Scala and static class members in Java. But perhaps your proposal, which changes the focus of the exercise to deal with semantic similarity between classes in Scala and Java instead is better to start off with as this is idiomatic Java. (Although static in Java is diffucult to understand for many.)  Maybe we should have some attribute that makes sense from a domain modelling point of vierw to keep static?

VictorWinberg:
Closing this PR since it shouldn't be merged, instead the question is available on issue #338.

PR #373, opened 2018-10-12
shuffle fixes

jos-rosenqvist-3812:


jos-rosenqvist-3812:
Ah, I didn't notice that PR #371 fixed the percentage thing first... But (hopefully?) this PR should merge cleanly even if that other PR is merged before this one.

PR #381, opened 2018-10-25
Add myself to the contributors list

wanecek:
Feels a bit lite patting my own shoulder, but... It's fun. Thanks for the offer @bjornregnell :smile: 

Thought: Would be pretty cool to generate this from the [contributors page](https://github.com/lunduniversity/introprog/graphs/contributors) with [GitHub's API](https://developer.github.com/v3/repos/#list-contributors) :) Either with Travis, or with GitHub's new Actions :thinking: 

**But** that could also lead to people getting their name in the book without them wanting to... Maybe not worth overengineering it

### DIFF
--- a/contributors.tex
+++ b/contributors.tex
@@ -1,7 +1,7 @@
 %  LIST OF CONTRIBUTORS to https://github.com/lunduniversity/introprog
-%    Please contact bjorn.regnell@cs.lth.se if you think you should be 
-%    on this list, or make a pull request with an update of file briefly 
-%    describing your contribtion in the commit text. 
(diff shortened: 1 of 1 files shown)


wanecek:
I also took the liberty of removing some trailing whitespace. Hope you don't mind - feel free to tell me otherwise, and I will drop the commit :+1:

bjornregnell:
There are many contributors that have contributed in many other ways than commits here at github, so automatic contrib list is neither feasible nor desirable: I want people to explicitly give consent e.g. by a PR procedure. Also "simple" fixes are valuable - e.g. even minor errors can be frustrating to a beginner programmer! Thanks for contributing!

PR #385, opened 2018-11-15
Remove unnecssary new before case class instantiation

wanecek:
As far as I can tell, the `new` keyword is redundant here, as we're instantiating a case class. Instead, it's confusing for those unsure of when to use `new` and when it can be omitted.

---

Found by @claracyon :tada:


wanecek:
After some more testing, it turned out that the `new` is indeed necessary for Scala to be other to distinguish between the (companion) `object` and the `case class`. TIL :)

PR #393, opened 2018-12-05
Avoid use of .recover in scalajava/hangman2 example

### LINKED ISSUE #390: [Question] Can Try in hangman.download of return a None?
In the finished example code in the idiomatic scala version of Hangman, [`hangman2.scala`](https://github.com/lunduniversity/introprog/blob/master/compendium/examples/scalajava/hangman2.scala), one apparent change from `hangman1.scala` is that [`hangman.download`](https://github.com/lunduniversity/introprog/blob/master/compendium/examples/scalajava/hangman2.scala#L23-L32) returns an `Option[String]`.

```scala
def download(address: String, coding: String): Option[String] = 
  scala.util.Try {
    import scala.io.Source.fromURL   
    val words = fromURL(address, coding).getLines.toVector
    val rnd = (math.random * words.size).toInt
    words(rnd)
  }.recover{ case e: Exception =>  
    println(s"Error: $e")
    "lackalänga"
  }.toOption
```

and the result is extracted through a foreach, i.e. [`download(url, codec).foreach(play)`.](https://github.com/lunduniversity/introprog/blob/master/compendium/examples/scalajava/hangman2.scala#L55)

As far as I see it, the possible outcomes are:

- The Try-block succeeds, and a string is returned (even if it's an empty string)
- The Try-block throws, and recovers with a fallback string (network request fails, file is empty (index overflow), etc)

Is there any scenario where `download` returns `None`? If so, how? And if not, would it make sense to `.get` the `Success`/`Failure`, instead of converting it to an `Option`, i.e. something along the lines of:

```scala
def download(address: String, coding: String): String =  // Changed to String
  scala.util.Try {
    import scala.io.Source.fromURL   
    val words = fromURL(address, coding).getLines.toVector
    val rnd = (math.random * words.size).toInt
    words(rnd)
  }.recover{ case e: Exception =>  
    println(s"Error: $e")
    "lackalänga"
  }.get // changed to .get
}
```

In the lab instructions, the following can be read:
> Låt metoden download returnera en Option[String] som i fallet att nedladdningen
misslyckas returnerar None.

However, as far as I can tell, that's not true - a failed download will either throw a `java.io.FileNotFoundException`, or, if the file is empty, throw an error when trying to access `words(rnd)`.

--- 

_I fully understand that there always are several ways to achieve a goal, often without one necessarily being better than the others. This seems to be especially true in Scala. I'm mostly curious to if I'm missing any of the outcomes of download, or if it's more idiomatic to use Option wherever possible._

wanecek:
## What

Let `download` directly convert the result of the Try-block to an option, forcing the callee to handle the error instead of providing a fallback case (and println) inside the download method.

**Note:** This _does_ introduce an inconsistency with `hangman1.scala` and `Hangman.java`, as these both `println` an error if such occurs. Not sure if that's relevant, I'll let you be the judge @bjornregnell.

## Why

Hopefully less confusing to others

## How

- Remove lines with `.recover{...}`
- Handle result with `Option#getOrElse`
- Update module .tex description (_This was the only place I could find that explicitly mentions the previous format, and I tried my best with some grepping of the `compendium/` directory, with words such as `download`, `läckalånga`, and `recover`. Would be great if someone could verify that I didn't miss anything this._)

---

This issue fixes #390

### DIFF
--- a/compendium/examples/scalajava/hangman2.scala
+++ b/compendium/examples/scalajava/hangman2.scala
@@ -26,9 +26,6 @@ object hangman {
       val words = fromURL(address, coding).getLines.toVector
       val rnd = (math.random * words.size).toInt
       words(rnd)
-    }.recover{ case e: Exception =>  
-      println(s"Error: $e")
-      "lackalänga"
     }.toOption
 
   def play(secret: String): Unit = {
@@ -52,7 +49,8 @@ object hangman {
   def main(args: Array[String] ): Unit = {
     if (args.length == 0) {
       val runeberg = "http://runeberg.org/words/ord.ortsnamn.posten"
-      download(runeberg, "ISO-8859-1").foreach(play)
+      val secret = download(runeberg, "ISO-8859-1").getOrElse("läckalånga")
+      play(secret)
     } else play(args((math.random * args.length).toInt))
   }  
 }
--- a/compendium/modules/w11-scalajava-exercise.tex
+++ b/compendium/modules/w11-scalajava-exercise.tex
@@ -111,7 +111,7 @@ import java.util.{HashSet => JHashSet};
 
 \item Låt metoden \code{download} returnera en \code{Option[String]} som i fallet att nedladdningen misslyckas returnerar \code{None}.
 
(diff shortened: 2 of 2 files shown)


bjornregnell:
It seems as if your main does not call play any more for zero args? Is it really correct or am I missing something=

PR #394, opened 2018-12-06
Fixed a regex to split commands in tabular labb

### LINKED ISSUE #392: w10_tabular input issue
The method fromUser() in Command splits input on space, which makes for example "filter OS Windows 10" filter for "Windows" and "10" instead of "Windows 10"

Sebergo:
Fixes #392 

Added a regex match instead of the simple split(' ') command in fromUser()

The command is however quite hard to understand so if we do not want to expose the stundents to regex it might be better to keep older version.


bjornregnell:
Sorry, regexp is not part of the course, so I'm closing this. Thanks anyway; perhaps this can be part of some advanced exercise or so.

bjornregnell:
Sorry regexp is not part of the course. Perhaps it can be part of some extra, advanced exercise in the future,

PR #410, opened 2019-10-01
"Primitiva Array" should be plural

AntiSC2:
This matches the text under it: "Primitiva arrayer".

An alternative could be: "Primitiv Array"

Signed-off-by: Jakob Sinclair <sinclair.jakob@mailbox.org>

### DIFF
--- a/slides/body/lect-w11-scalajava.tex
+++ b/slides/body/lect-w11-scalajava.tex
@@ -617,7 +617,7 @@ object Person {
 
 \Subsection{Array}
 
-\begin{Slide}{Repetition: Primitiva Array i JVM}
+\begin{Slide}{Repetition: Den primitiva typen Array i JVM}
 \begin{itemize}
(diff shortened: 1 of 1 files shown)


jos-rosenqvist-3812:
I think this might be correct already, having the meaning of "[den] primitiva [typen] Array". But "primitiva arrayer" would also make sense (though perhaps with a lowercase a).

bjornregnell:
@AntiSC2 I'm inclined to keep it as is to have the type Array explicitly shown. Could you go with @jos-rosenqvist-3812 solution by expanding the implicit ''Den primitiva typen Array i JVM''

AntiSC2:
> @AntiSC2 I'm inclined to keep it as is to have the type Array explicitly shown. Could you go with @jos-rosenqvist-3812 solution by expanding the implicit ''Den primitiva typen Array i JVM''

@bjornregnell, it makes sense to me so will do that

PR #412, opened 2019-10-02
Fix incorrect indexing in week 6 slides

nichobi:
The thirteenth element of xs is at index 12, not 13.


bjornregnell:
Many thanks! If you haven't done so already, you are welcome to also make a PR on https://github.com/lunduniversity/introprog/blob/master/contributors.tex adding your name in the list in alphabetic order on first name, respecting punctuation and formatting as is.

PR #413, opened 2019-10-04
Fix references to old library code

TheJarles:
The text was referring to String-variables in introprog.PixelWindow.Event, which have been refactored to Int-variables (introprog-0.1.1). Also along with this change the REPL-output of introprog.examples.TestPixelWindow changed.


bjornregnell:
Many thanks! If you haven't done so already, you are welcome to also make a PR on https://github.com/lunduniversity/introprog/blob/master/contributors.tex adding your name in the list in alphabetic order on first name, respecting punctuation and formatting as is.

PR #415, opened 2019-10-06
Fix/small update

MiroBro:
fix code typo in exercise w06 7.c.)
add descriptive detail about methods mentioned in exercise w05 7.e.)
(Added this because when I was a teaching assistant during the week
several students expressed confusion regarding the methods in
this final exercise. Thought maybe small added details may help?)


bjornregnell:
@MiroBro Many thanks! Next time, try to make a more specific commit message to help understand what the fix is about and where and perhaps why, but keep it short... Not always easy but if a good msg can be made it helps a lot when browsing the list of all PR:s in history...

PR #427, opened 2019-11-13
javatext overhaul

### LINKED ISSUE #423: javatext: adjust mandatory reqts and focus/reduce the effort
@jos-rosenqvist-3812 Thanks for reminding me about this issue!

Javatext currently has these problems:
* students spend too much time on things that don't matter so much
* too many are making everything static

Solution idé:
* update mandatory requirements to make it clear and less cumbersome
* focus the effort on illustrating some important differences between Scala and Java and training the idiomatic java class ''anatomy' with private atributes and constructors and getters setters etc.

Discussion:
* Use this issue thread if discussions on actual solutions are needed.

jos-rosenqvist-3812:
Closes #423.


bjornregnell:
fixes issue #423

bjornregnell:
@kattaroten Could you review this PR after reading the issue #423 and give your views on the change in this thread? If you have the time before Sunday, it would be great and then I'll check this comment thread on Sunday and decide on a merge.

kattaroten:
This seems like a thoughtful and elegant solution to the issues the lab currently has. I think it definitely should be adopted as the new description for the lab even if I do have some minor issues with it.

My main gripe is really the removal pf the IO-parts. I do agree that it does not feel appropriate for this lab but I also think it's an important realisation about how to interact with the file system. Where in the course would it be approached if removed from this lab? There are several labs that have it as an optional task but none where it is central.

I am a bit worried about the switch-case used in that way. It doesn't look like a java idiom to my (admittedly not very experienced) eyes. 

In the interest of further preventing confusion I would suggest changing the name of UserInterface to something else to avoid mistaking it for or seeing a false relatedness to java interfaces. Perhaps UserView or something similar. It is a peripheral issue however but I wanted to include it just as a thought.

bjornregnell:
Thanks @kattaroten ! I'm not that worried about the IO parts being removed, as it is part of words and the Bank project, and the "real stuff" about IO is actually streaming and nio and that is beyond this course. I think UserInterface is a good name (or maybe TerminalUserInterface but that is too long) and I'm not worried about the link to the Java keyword; they are actually interfaces - both api and UI and GUI are interfaces, so I see no problem here.  The java switch - case looks good to me. 
@jos-rosenqvist-3812 I'm planning to merge this soon, if you don't have any other objections? Improvements, if needed, can come as PR:s after merge,

jos-rosenqvist-3812:
> @jos-rosenqvist-3812 I'm planning to merge this soon, if you don't have any other objections/proposals. Any improvements, if needed, can come as PR:s after merge.

I don't have any objections.

Issue #421, opened 2019-10-15
improve BlockBattle instruction and doc of methods


bjornregnell:
Explain and document given code. Consider rename eraseBlocks to redraw blocks. BackgroundColorAtDepth etc, Comments welcome.

OliverPerssonB:
One economical edit is simply adding a comment line, like the on in update bellow, saying something like "Use to erase old points". This can also be done in the colorAtDepth method, giving more hints to the purpose of the function.

bjornregnell:
@OliverPerssonB  Would you be prepared to makes this simple fix in a PR?

OliverPerssonB:
Yes, I will fix it during the week.

PR #431, opened 2019-11-25
blockBattle: added comments, see issue #421

OliverPerssonB:
These two commented functions where often left unimplemented in a working solution by students. Hopefully, these comments are strong enough hints to get them rolling and using these functions.


bjornregnell:
najs!!

PR #448, opened 2019-12-29
Revert "Fix implied typos"

bjornregnell:
over-populated or over populated are two words in English!?


bjornregnell:
@elliotbrack As there are two changes in the same commit of which I after merge realized that I only wanted to merge one of the this became a revert. Could you get back with a new PR that fixes BankKonto -> Bankkonto, which indeed is one word in Swedish? (Under populated in US-english and Under-populated in Britt.-eng. are actually two words...)

elliotbrack:
Are you sure?

https://www.merriam-webster.com/dictionary/underpopulated

PR #467, opened 2020-08-12
Adding three dots

benkibejs:
I think i found the simplest solution and something that would still make people remember the example. You can always ask the class about what is missing and see if they figure it out. The example of gender as a binary thing is a flawed way to describe human expression, behaviour and interaction.


bjornregnell:
Ok, let's see how students react to this!

bjornregnell:
Thanks for all your work @benkibejs on issues etc. You are welcome to come with a PR that adds you here if you want: 
https://github.com/lunduniversity/introprog/blob/master/contributors.tex
Note: Alphabetical on first name...

PR #478, opened 2020-09-10
Fix #475 added missing definition for r

### LINKED ISSUE #475: Typo in assignment 18, chapter 2.2.3
There's a typo on the second line in the listed code. It states: `val sekvens = r.toArray` and r is never defined.

wudse20:
Fix for #475, the missing defintion for r has been added, based on the answer in chapter L.2.3.

### DIFF
--- a/compendium/modules/w02-programs-exercise.tex
+++ b/compendium/modules/w02-programs-exercise.tex
@@ -1355,7 +1355,7 @@ res3: Double = 288.6589258453995
 
 \begin{REPL}
 scala> val intervall = (1 to Int.MaxValue by 2)
-scala> val sekvens = r.toArray
+scala> val sekvens = intervall.toArray
 \end{REPL}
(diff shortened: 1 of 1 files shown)


emmanuelkring:
I think the original mistake here was to write `r` instead of `intervall`. It can be fixed in the following way:
> scala> val intervall = (1 to Int.MaxValue by 2)
scala> val sekvens = intervall.toArray

wudse20:
This PR should cover your fix, which is better.

jos-rosenqvist-3812 (approved):
The commit history of this pull request would look nicer if the commits were squashed into one and the merge commit was removed: https://stackoverflow.com/questions/30561895/removing-merge-commits-in-git/30565914

However, this can be tricky to do if you don't already have a lot of experience with git, so alternatively we could merge the pull request as-is. Please let us know if this is something you would like to do before the pull request gets merged.

wudse20:
The merge commit should now be removed.

PR #482, opened 2020-09-18
w02: Fix typos

sadphi:
Fixed some typos in the lecture and exercise section in w02


jos-rosenqvist-3812 (approved):
The commit history of this pull request would look nicer if the merge commits were removed: https://stackoverflow.com/questions/30561895/removing-merge-commits-in-git/30565914

However, this can be tricky to do if you don't already have a lot of experience with git, so alternatively we could merge the pull request as-is. Please let us know if this is something you would like to do before the pull request gets merged.

sadphi:
Hi Jos! I tried following the solution from the link, but it looks like I just made it worse... So you were right about it being tricky 🙄. 

At this stage I think it's better if I redo the commits from a clean fork.

PR #503, opened 2020-10-18
Fix #502 removed deprecated use of prepend in Chapters 7.2.3 and L.7.3

### LINKED ISSUE #502: Deprecated code in assignment 20, chapter 7.2.3 and L.7.3
In assignment 20, chapter 7.2.3 and L.7.3, the method: <code>scala.collection.mutable.ListBuffer.prepend</code> is used which is deprecated since Scala 2.13.0, with multiple arguments, and replaced by <code>scala.collection.mutable.ListBuffer.prependAll</code>.

wudse20:
Fixes issue #502 and removes the deprecated uses of prepend and swaps it with prependAll.


bjornregnell:
Good catch! Thanks!

PR #512, opened 2020-11-27
javatext: Flush System.out after print

jos-rosenqvist-3812:
`System.out` automatically flushes when a newline is printed, but we can't rely on automatic flushing when not printing a newline.


jos-rosenqvist-3812:
Alternatively, if we don't want students to have to think about what flushing is, we could change this to simply use `println` instead of `print`. But perhaps encountering the concept of flushing might not be a bad thing?

PR #513, opened 2020-11-30
Update w12 slides for 2020

jos-rosenqvist-3812:
I especially don't want students to get confused by the link http://cs.lth.se/pgk/sam which contains data from 2019.


bjornregnell:
I have now removed the old sam.html on filedamin

PR #521, opened 2021-01-21
Remove exercise 7 task 3 (abstract collections quiz)

### LINKED ISSUE #497: Iterable occurs twice in exercise 7 task 3
image

635507f replaced usages of Traversable with Iterable, but a simple replacement doesn't make much sense in this case, since we end up with two Iterable. What should be done about this task? Having only two items feels like a bit too little for this "matchup" type of task...

jos-rosenqvist-3812:
Fixes #497. But unfortunately this makes the "mutant task" be task 4 instead of task 5 :(


bjornregnell:
Is the 5 -> 4 a problem? Is the hard-coded number referenced somewhere, and in that case where?

jos-rosenqvist-3812:
It's not an actual problem. The task just has a lot of references to the number 5, so I thought it was fitting that it was task 5 as well.

bjornregnell:
Aha. Fem fem fem.   https://youtu.be/dpdOUEe9mm4
:laughing:

PR #535, opened 2021-06-13
Migrate exercisew06 patterns to Scala 3 fix #533

### LINKED ISSUE #533: Migrate exercise w06 patterns to Scala 3
I've made a fork where I have begun migrating code to Scala 3, however...

In exercise 3.b the student is instructed to press tab twice. Using the old scala 2 repl, this would have given the right outcome whilst the new `scala3-repl` doesn't give the right outcome. The exercise instructions should therefore be rewritten if the `scala3-repl` is to be used by students.

In exercise 3.c, the returned value is no longer `Some((100, true))`, instead the value `Gurka(100, true)` is returned when working in the `scala3-repl`. I guess the solution should be changed to match the new outcome?

In exercise 4 the student is instructed to enter `:pa` to the repl to enter paste mode, while this doesn't seem to be a feature of the `scala3-repl`. Also, it seems that the behaviour of `sealed trait` in `scala3-repl` has changed from the old repl. No longer are you able to extend a sealed trait defined in the repl (bug or feature?): 
```
scala> sealed trait Färg
// defined trait Färg
scala> case object Spader extends Färg
1 |case object Spader extends Färg
  |            ^
  |            Cannot extend sealed trait Färg in a different source file
```
It works if you change `sealed trait` to just `trait`, if this is a desired solution then the instruction of the task must be rewritten.
If you change to just `trait` however, subtask 4.b is no longer relevant, since then there is no warning of a case missing. Also subtask 4.d loses its purpose of explaining the keyword `sealed`.

alfredgrip:
All code should be migrated to Scala 3.

Left some TODOs that needs to be reviewed.

Left a commented out skeleton to a possible `enum`-task.


bjornregnell:
It just struck me that unapply has changes significantly in Scala 3 and it is "option-less":
```
scala> Gurka.unapply                                                            
val res6: Gurka => Gurka = Lambda$1805/0x0000000840836840@7f086b45
```
See the gory details here:
https://dotty.epfl.ch/docs/reference/changed-features/pattern-matching.html
So in the patterns exercise we either need to find a way to explain and let the student investigate this without getting into too many details, or simply leave the unapply story to lectures and don't exercise it. What do you think?

alfredgrip:
> 
> It just struck me that unapply has changes significantly in Scala 3 and it is "option-less":
> 
I did notice this when I went through all exercises. All I changed when noticing this was the key (facit) to one subtask which was affected by this.

I think that the unapply method doesn't bring that much to the table for the student. Leaving it to only lectures would be better. Personally when I attended this course I can't remember the unapply method ever being used in a meaningful way, so maybe it's not necessary for students to exercise it, but instead live it to lectures. However that's just my opinion.

bjornregnell:
Ok, so remove the TAB-TAB stuff and all stuff related to unapply.

bjornregnell:
Also I think we should have a subtask after the sealed trait tasks, that also ask for how this can be done with enum. Could you fix that (if y9ou also think it is a good idea)?  I guess I will introduce enums earlier than week 6 but it is always nice to exercise what you have learned previous weeks and enums are prefect for patternmatching. `case Färg.Spader => ... ` etc

alfredgrip:
I have now added an `enum` subtask and removed things related to unapply. Please do review the task I added!

bjornregnell:
The translation "förlänger" of "extends" has not been used before. I prefer to use "ärver" as in *inherits*. Please fix that.

compendium/modules/w06-patterns-exercise.tex
```diff
@@ -281,6 +302,49 @@
 \QUESTEND
 
 
+\WHAT{Mönstermatcha med nyttan av \code{enum}.}
```
bjornregnell, line 305: Remove "nyttan av" as the sentence reads a bit strange. I suggest instead the simpler title: "Mönstermatcha enumeration"

compendium/modules/w06-patterns-exercise.tex
```diff
@@ -281,6 +302,49 @@
...
+\TODO %Se över denna
+\SubtaskSolved
+\begin{Code}
+def parafärg(f: Färg): Färg = f match
+  case Färg.Spader  => Färg.Klöver
+  case Färg.Hjärter => Färg.Ruter
+  case Färg.Ruter   => Färg.Hjärter
+  case Färg.Klöver  => Färg.Spader
+\end{Code}
+
```
bjornregnell, line 340: Add a comment that you can use import to skip `Färg.` but it is often better to keep it to avoid name clashes. Compare if you had several color Red one for cards and one for Pixels...

compendium/modules/w06-patterns-exercise.tex
```diff
@@ -940,6 +992,7 @@
 
 \Task  \what~   Om man överskuggar den befintliga metoden \code{equals} så kommer metoden \code{==} att fungera annorlunda. Man kan då själv åstadkomma innehållslikhet i stället för referenslikhet. Vi börjar att studera den befintliga equals med referenslikhet.
 
+\TODO %tab två gånger fungerar inte i scala3-repl, issue #536
```
bjornregnell, line 995: Remove/convert TAB stuff to something that works in Scala 3

bjornregnell (changes requested):

compendium/modules/w06-patterns-exercise.tex
```diff
@@ -337,20 +335,23 @@
...
+Anta också att vi nu till vårt program vill importera ytterligare en röd färg för kulörerna hjärter och ruter, denna också namngiven \code{Red}. I detta scenario hade det uppstått en namnkrock då \code{Red} redan är definerad så importeringen hade ej kunnat ske.
 
 \SubtaskSolved
-\TODO %Se över denna
-Vid mönstermatchning så fungerar \code{sealed trait} ihop med \code{case}-objekt i praktiken likadant som att använda sig av \code{enum}. Vi såg att i deluppgift \ref{subtask:match-forgetcase} så varnade REPL redan vid kompilering att denna funktionen inte var uttömd \Eng{exhaustive}. Detta gäller även vid användning av \code{enum} också.
+Vid mönstermatchning så fungerar \code{sealed trait} ihop med \code{case}-objekt i praktiken likadant som att använda sig av \code{enum}. 
+Vi såg att i deluppgift \ref{task:match-sealedtrait}\ref{subtask:match-forgetcase} så varnade REPL redan vid kompilering att denna funktionen inte var uttömd \Eng{exhaustive}. Detta gäller även vid användning av \code{enum} också.
```
bjornregnell, line 344: change "denna funktionen inte var uttömd" till "denna matchning inte var uttömmande"  
remove "också" (redundant to "även")

bjornregnell (changes requested):

bjornregnell (commented):
LGTM

PR #550, opened 2021-06-20
Migrate exercise w03 functions to Scala 3 fix #542

### LINKED ISSUE #542: Migrate exercise w03 functions to Scala 3

NiklasSanden:
I added some \TODOs in the file:
1. There are two files in compendium/generated/ that I would like to change. Where should I look?
2. "Uppgift 6" is no longer as relevant since the underscore can be omitted in all of the cases described in the exercise. Should I try to change it? Currently, the answers include both the correct answer for Scala 3 and the previous results.
3. "Uppgift 15" uses lastException and :javap which don't seem to be available in the Scala 3 REPL. For now I changed the code and answer to not include them. Since the REPL prints out the stack trace along with the exception I don't think lastException is necessary. Is there another way to show the machine code inside the REPL?

### DIFF
--- a/compendium/examples/AliensOnEarth.scala
+++ b/compendium/examples/AliensOnEarth.scala
@@ -28,8 +28,7 @@ object AliensOnEarth:
       val points = playGame(alien = name)
       if points > 0 then println(s"Congratulations $name! :)")
       println(s"You got $points points.")
-    catch { case e: Exception =>
+    catch case e: Exception =>
       println(s"Game over. The Earth was hit by an asteroid. :(")
       if isAnswerYes("Do you want to trace the asteroid?") then
         e.printStackTrace()
-    }
--- a/compendium/examples/AliensOnEarth.scala
+++ b/compendium/examples/AliensOnEarth.scala
@@ -1,6 +1,6 @@
 object AliensOnEarth:
   def readChoice(msg: String, options: Vector[String]): String =
-    options.indices.foreach(i => println(i + ": " + options(i)))
+    options.indices.foreach(i => println(s"$i: ${options(i)}"))
     val selected = io.StdIn.readLine(msg).toInt
     options(selected)
 
--- a/compendium/modules/w03-functions-exercise.tex
+++ b/compendium/modules/w03-functions-exercise.tex
@@ -227,7 +227,6 @@ scala> val xs = Vector(3, 4, 5)
 \end{REPL}
 \noindent Para ihop nedan uttryck till vänster med det uttryck till höger som har samma värde. Om du undrar något, testa uttrycken och olika varianter av dem i REPL.
 
-\TODO Hur kan man ändra på dessa?
 \begin{ConceptConnections}
 \input{generated/quiz-w03-yield-map-taskrows-generated.tex}
 \end{ConceptConnections}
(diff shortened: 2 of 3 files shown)


bjornregnell:
Thanks for your work!
1. What type of change are you looking for?
2. I think task 6 in w03 is still valid as you sometimes still need understreck (see new slides that week https://github.com/lunduniversity/introprog/commit/186b1bb9b23ea4618f3858d8b0fe11205ea58b0f)  But maybe som sub-tasks should be adopted to the new situation, by starting off without underscore as the new normal and then have just one subtask by the end that illustrates when underscore is really needed. Another way is to start with underscore but in an early sub-task say that it most often kan be skipped and let them try that. What do you think?
3. You can run javap in terminal if you jave a LDK install but I'm not shure how important that is to include in the exercise. You can try it yourself by compiling in terminal and doing `javap -c *.class` to see the byte code. hat is if if you install the scala binaries from the scala 3 release manually and put scalac and scala on your path (still no good way of installing the binaries compared to scala 2 :(  ), or if you use sbt you find the class-files in target: `javap -c target/scala-3.0.0/classes/*.class`
And yes lastException can be skipped. (sometimes the full stack was elided in Scala 2, but seems as you get to see the relevant stuff better in scala 3)

NiklasSanden:
1. I was going to remove the parentheses in the for loop on the first line in generated/quiz-w03-yield-map-taskrows-generated.tex and in the answer, generated/quiz-w03-yield-map-solurows-generated.tex (which seems to be the preferred style used in the theory section). My understanding is that changes to these files directly would be overwritten when rebuilding.
2. I was trying to find more examples of where an under score was needed. [This](https://dotty.epfl.ch/docs/reference/changed-features/eta-expansion-spec.html#rules) seems to suggest to not use under scores at all for this purpose and prefers in our example:
val b = () => a() - or - val b: () => Int = a
The latter is similar to the old task but now for a more specific case, so I will change it to include this. Should the slides perhaps suggest one of the versions above instead of the under score (which, if I understood it correctly, should be or will be deprecated depending on where on that page you read)?

bjornregnell:
Good that you found it! Yes all generated stuff will be overwritten on each full build. 

Regarding point 2: Yes, that page says "The syntax m _ is no longer needed and will be deprecated in the future." so we better avoid it completely (but there is no way to auto-eta-expand a def without parenthesis other than using a thunk if the underscore is deprecated, but that's ok ). I'll think about if I should keep underscore for explicit eta-expansion on lectures as students might encounter Scala 2 code with underscore in this position. I'm also considering to introduce the term "Thunk" (in swedish?)  for () => bla. https://en.wikipedia.org/wiki/Thunk

Anyway: make the subtask as you think best fit the Scala 3 situation post deprecation and I'll review it when ready. Ok?

NiklasSanden:
Sounds good!

NiklasSanden:
I changed the task to have a slightly different focus to include everything we mentioned. It no longer has any relation to the previous task (like it did before with "öka"). If it was changed too much I can instead just make subtasks e and f replace the old d and e (with some function defined in a that takes no parameters).

bjornregnell:
Thanks! I'll review your new proposals as soon as I'm done with some other urgent stuff.

bjornregnell (approved):

PR #554, opened 2021-06-25
Migrate exercise w07 to Scala 3

elsaogestad:
I have now changed all the exercises from week 7 to Scala 3. But I have a few questions/remarks:

1. Exercise 25 is about WrappedArray, but it doesn't seem to exists anymore. Should it be removed completely or something else?
2. Exercise 9, 18, 22, and 27 are all about looking up in Scala doc or "snabbref". That doesn't seem to be updated in Scala 3, but I haven't found any problem using it. I think that I will leave it for now.

### DIFF
--- a/compendium/generated/quiz-w07-seq-methods-solurows-generated.tex
+++ b/compendium/generated/quiz-w07-seq-methods-solurows-generated.tex
@@ -1,8 +1,8 @@
   \code|x +: xs         | & 1 & ~~\Large$\leadsto$~~ &  G & \code|Vector(0, 1, 2, 3)                      | \\ 
-  \code|xs +: x         | & 2 & ~~\Large$\leadsto$~~ &  L & \code|error: value +: is not a member of Int  | \\ 
+  \code|xs +: x         | & 2 & ~~\Large$\leadsto$~~ &  L & \code|value +: is not a member of Int         | \\ 
   \code|xs :+ x         | & 3 & ~~\Large$\leadsto$~~ &  J & \code|Vector(1, 2, 3, 0)                      | \\ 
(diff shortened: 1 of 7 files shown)


bjornregnell:
1. Yes, `WrappedArray` is replaced by `ArraySeq`, an immutable sequence backed by a private primitive Array. Scala 3 use the same standardlibrary as 2.13 which also includes ArraySeq. IArray only existst in Scala 3 not in 2.13. You can read about ArraySeq here: https://docs.scala-lang.org/overviews/core/collections-migration-213.html   
In Scala 3 there is also `IArray` which uses the new zero-overhead abstraction Scala 3 feature called opaque types.  You can read about IArray here: https://dotty.epfl.ch/api/scala/IArray$.html and a lot of technical details here https://medium.com/scala-3/scala-3-safer-pattern-matching-with-matchable-f0396430ded6  and here https://dotty.epfl.ch/docs/reference/other-new-features/matchable.html . I have not yet decided if and if so in what week I will lecture about opaque types, but it is a performance thing and not central to this course. 
So I suggest that you adapt the existing exercise to use `ArraySeq`. Or du you see any problems with that? (I havent checked the exercises yet.)

2. I haven't started to migrate the quickref, but everything that relates to collections should be the same so keep current exercises if they still work.

bjornregnell:
Do you need to change anything more in the light of my answers regarding ArraySeq or is this ready for merge?

elsaogestad:
Yes, I haven't had the time to change exercise 25 so I will do that today. Everything should be done after that!

bjornregnell:
:+1:

elsaogestad:
I have now changed exercise 25. I left some links to the old docs as the new is not updated.

bjornregnell (approved):

PR #565, opened 2021-08-08
Add w13_photo with ImageFilter.scala

theolundqvist:
Contains old scalalib update. Reference to git/introprog-scalalib will overwrite this later?


bjornregnell:
I think we should skip the files that are >= 100kB, as big binaries don't play well with git.

PR #566, opened 2021-08-13
w05: Fix exercise quizzes and improve solutions

sadphi:
I just discovered that the quizzes in exercise w05 were reset to the version pre-migration, which was due to me incorrectly editing the generated `.tex` files in `compendium/generated` instead of `quiz/QuizData.scala`. This should be fixed in this PR, but I can't build the project as I currently have issues with Scala and sbt, so I haven't been able to check it myself. But according to the Github Actions run it should be fine. I also tried running `sbt build` on another PC that is running Windows, but the build failed there due to Windows missing the program `tail`.

### DIFF
--- a/compendium/modules/w05-classes-exercise.tex
+++ b/compendium/modules/w05-classes-exercise.tex
@@ -89,7 +89,7 @@ class  Punkt       { var x = 3; var y = 2 }
 ~\\ \emph{fel} & \emph{typ} & \emph{förklaring} \\\hline
 
 \code|value is not a member of object|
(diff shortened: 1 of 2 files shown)


bjornregnell:
Very good that you spotted this! However it seems as if some of the behaviour is not Scala 3 but Scala 2; have you tested this with Scala 3? For example `MutablePt()` should compile and new is not needed in Scala 3 etc.  (For details on universal apply methods see https://docs.scala-lang.org/scala3/reference/other-new-features/creator-applications.html  )

bjornregnell:
Let's discuss this live on Discord before further actions.

bjornregnell:
Let's move strict equality to in-depth exercises.

sadphi:
I've now moved the exercise on strict equality to the in-depth exercises, and revised it to just be about analysing the behaviour.

PR #568, opened 2021-08-19
w13-photo assignment

theolundqvist:
- update project text
- add starting point


theolundqvist:
@bjornregnell Don't forget to 
- update scalalib to 1.2.0 and 
- update scalalib docs

PR #582, opened 2021-08-22
Add name to contributors.

theolundqvist:
Too bad I was late :(


bjornregnell:
Will resend so you get in there. :)

PR #584, opened 2021-08-27
web/tools: Fix typo

sadphi:
Quick typo fix. I didn't update the html file with `build.sh` as the output file had some other unexpected changes.


bjornregnell:
good catch! now published

bjornregnell:
it sometimes takes some minutes until the typo 3 backend refreshes and the javascript hack pulls from github, but hopefully its soon visible after reload with cleared cache (Ctrl+F5 in Firefox).

sadphi:
Great!

PR #591, opened 2021-09-14
updating example code for clarity

M0R60TH:
Changing from a print to println to avoid collision status text in sbt since it prints directly on the line. This is also concurrent with function usage in other parts of the code.


bjornregnell:
Looks good! Why did you close it? (It needs to be open for me to merge it)

M0R60TH:
Closed it because I wanted to open an issue for discussion of the topic.

bjornregnell:
aha, yeah good to connect to issue, but you can leave a PR open and mention the issue number here and they will be connected. So open this again and mention the issue with a #

M0R60TH:
This is discussed in #592

PR #593, opened 2021-09-15
Remove duplicate chapters

M0R60TH:
In the current release chapter 3.1.15 is an exact copy of 3.1.14. This edit removes the duplicity.

PR #594, opened 2021-09-15
Removing duplicate chapters, an update on #593

M0R60TH:
In the current release chapter 3.1.15 is an exact copy of 3.1.14. This edit removes the duplicity.

### DIFF
--- a/slides/body/lect-w03-functions.tex
+++ b/slides/body/lect-w03-functions.tex
@@ -413,7 +413,6 @@ val res1: Int = 40
 \end{REPL}
 \end{Slide}
 
-
 \begin{Slide}{Applicera funktioner på element i samlingar med \texttt{map}}\SlideFontSmall
 \begin{Code}
 def öka(x: Int) = x + 1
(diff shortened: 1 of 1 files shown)


Husenap:
they're actually not identical, the first subsection asks the reader what the result will be, which is then revealed in the next subsection

```scala
val res0: ???   // vad blir resultatet?
```

```scala
val res0: scala.collection.immutable.Vector[Int] = Vector(2, 3, 4)
```

This makes perfect sense from a "slides-perspective", but it's a little bit awkward from a "book-perspective"
🤔

M0R60TH:
Hm, I see. Thanks for clarifying.

bjornregnell:
It is possible to exclude slides from compendium by surrounding them with `\ifkompendium\else` and `\fi` -- look in the code for examples. But in this case I think we should keep both in the compendium as the comment indicates that the reader should think about the result marked with ???, as an exercise for the reader. So I'm closing this as a "won't fix".

PR #595, opened 2021-09-21
Cleanup chapter 4

Husenap:
* Removed duplicates on the chapter4 plan
* Renamed `HighScore` to `Highscore` in figures
* Fixed a few typos
* Updated solution for task 4.12


bjornregnell:
We should keep _1 and _2 as I think the apply method is less efficient on large tuples and the normal getters with underscore are not "old" syntax" but the idiomatic way to access tuple parts as I understand it.

bjornregnell:
Also, there is not any rule on consistency for large or small letters on object names; both are seen in Scala code, and capitalized inits are perhaps more often used for companion objexts, i.e. when there is a class with the same name (but not only). But we can have it as you propose, but I am a hesitant as with a lower case init you see that it is a reference name to a single value similar to a val.

Husenap:
> We should keep _1 and _2 as I think the apply method is less efficient on large tuples and the normal getters with underscore are not "old" syntax" but the idiomatic way to access tuple parts as I understand it.

The online Scala 3 book doesn't mention the `_n` syntax and uses `(n)`.
The Programming in Scala book uses the `(n)` syntax and says:
"prior to Scala 3, you accessed the elements of a tuple using one-based field names, such as _1, _2."

That's why I made the conclusion that `(n)` might be more fitting for Scala 3.

Husenap:
> Also, there is not any rule on consistency for large or small letters on object names; both are seen in Scala code, and capitalized inits are perhaps more often used for companion objexts, i.e. when there is a class with the same name (but not only).

You're right, there isn't any rule for this, I only made it consistent with the other examples in this specific chapter, for example `object Highscore`.
But I see that there are tons of other examples in other chapters that also have a lowercase init.

> but I am a hesitant as with a lower case init you see that it is a reference name to a single value similar to a val.

This is a good point, so perhaps it's not worth making this change.

bjornregnell:
Interesting that apply is the new idiom in Alvin Alexander's et al  Scala 3 book. I guess that book takes on some responsibility to set a new standard... Can ju link to the exact page where _1 is not there but apply(0) is ? (Could find it and there is no easy search...)

Husenap:
This is the page where I saw it: 
https://docs.scala-lang.org/scala3/book/taste-collections.html#tuples

bjornregnell:
When thinking more of it I think it is bad to not show _1 as I believe tons of Scala code use it and it is a bit cryptic until you get it so better tell about it. I also like that they are methods, albeit with special names. But I see their point of sticking to zero-based indexing. But am not currently prepared to adopt this as this will impact so much of the teaching material. But it would be ok to mention somewhere that apply is also possible in scala 3.

bjornregnell:
BTW: you have committed a generated file. This will be overwritten by the next build so better fix its source...

Husenap:
> BTW: you have committed a generated file. This will be overwritten by the next build so better fix its source...

Yes I noticed that(oops), I've fixed it locally in `Plan.scala`

PR #596, opened 2021-09-21
Fix chapter 3 typos

Husenap:
* Fixed a couple of typos in lab preparations

### DIFF
--- a/compendium/modules/w03-functions-lab.tex
+++ b/compendium/modules/w03-functions-lab.tex
@@ -7,8 +7,8 @@
 \end{Goals}
 
 \begin{Preparations}
-\item Gör övning övning \texttt{\ExeWeekTHREE} och repetera övning \texttt{\ExeWeekTWO}.
-\item Läs appendix~\ref{appendix:terminal} och ~\ref{appendix:compile}.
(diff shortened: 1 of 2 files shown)


bjornregnell:
Thanks for looking into this. Yes if we deprecate underscore then we need to go through the whole compendium and slides as I think they might be in several places... Would be nice if you could investigate the impact before going with a thunk.
I think we should properly introduce the term "thunk" in the `function` week and make a separate slide on that.
https://en.wikipedia.org/wiki/Thunk

Husenap:
How about I leave it out of this PR and only keep the typo fixes, then I can create an issue where we can discuss it further?

bjornregnell:
Yes, good to separate typos and the longer stuff!

bjornregnell:
There are alot of formatting changes, so it is easy to miss other changes when reviewing. Are all the indents from the original file?  Usually I dont use indents in the latex code as there are often unwanted side effects of those...

bjornregnell:
Ah. I was looking at an old page - now when refreshing in browser the changes look nicer and only a few. Cool!

bjornregnell:
Is this ready for merge?

bjornregnell:
Looks Good To Me.

Issue #601, opened 2021-09-24
Create exercise that train how to rewrite `do while` to `while do`


M0R60TH:
The 13th part of the lab for week 4 tells the student to try out the "do while" loop. Since this is now deprecated, should this be removed/changed in some way?

![image](https://user-images.githubusercontent.com/55506884/134669524-215ac78c-4115-4145-bead-62c6f91289f9.png)

bjornregnell:
Either it should be deleted or help people realize how flexible an ordinary while do can be used in line with the examples here:
https://docs.scala-lang.org/scala3/reference/dropped-features/do-while.html
What do you think? Shoul'd we delete it or is there an opportunity to show "do-while-imitating" while-do? And if the latter, how?

M0R60TH:
I'd say that it could probably be removed, for the same reasons that the folk behind scala give for dropping it;

_do-while is used relatively rarely and it can be expressed faithfully using just while. So there seems to be little point in having it as a separate syntax construct._

bjornregnell:
Yeah. But perhaps a new exercise in the "going deeper" part? With a link to the "dropped-feature" page above. Saying something that "In many languages there are a backwards do while that evaluates the condition after the first rund .... but while do can also do that like so ..." and some exercise with a very small algorithm where it still makes sense.
PR welcome...

bjornregnell:
But that maybe should go in an earlier week where we introduce while. Or perhaps in a later week when we have more algorithms. This needs some consideration... But again its not super-important :)

bjornregnell:
Anyhow it should be removed from the basic exercises, so you can come with PR on that removal if you like?

M0R60TH:
I'll do a PR for the remove and leave you to figure out if and where to add a possible new exercise 🙃

bjornregnell:
Ok let's keep this issue open even if we delete the deprecated stuff.

PR #602, opened 2021-09-24
removing deprecated part

M0R60TH:
do while is deprecated as of scala 3, this part shoudl therefore be removed


M0R60TH:
this is in accordance with the discussion in #601

PR #617, opened 2021-11-17
Removed duplicate code from example

Roborhugo:
Removed a second addition test identical to the first.


bjornregnell:
Thanks!! Good catch!

PR #656, opened 2022-09-07
Update lect-w01-intro.tex to improve clarity

HampSwe:
As "scala-cli" has "run" as a command, a beginner who has used the command "scala-cli run ." previous to reading this might mistakenly believe that "scala" also has the command "run". However, it doesn’t: "run" only refers to the main function in this example. It would therefore be more pedagogical if the main function was called something different (like "hi").


bjornregnell:
Thanks! Good idea! Next year hopefully scala-cli will be the new scala, but it still can be mixed up with the name of the sub-command conceptually.

PR #661, opened 2022-09-19
Add scala-cli workspaces

oliverlevay:
Port w03->w10 workspaces to scala-cli in a new folder called `workspaces_scala-cli`, all can compile and run (with a lot of unimplemented exceptions of course).

Mostly it involves removing build.sbt and project folders and adding
`//> using lib "se.lth.cs::introprog::1.3.1"` at appropriate places.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ about/**/.vscode
 *~      # emacs
 
 .idea
+.vscode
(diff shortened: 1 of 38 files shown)


bjornregnell:
Great! I'm considering if this should be the default workspace; remind me on our lunch meetings to discuss this if time permits, @oliverlevay

bjornregnell:
What do you think: Perhap it would be better to have the sbt stuff in a subdir called sbt for symetry reasons? But then we need to check if that break things first,

bjornregnell:
I'm merging this for now as it does not impact anything, and then we can consider moving the existing sbt-stuff in a subdir later.

Issue #663, opened 2022-09-20
Consistent language regarding "Uniform Access/Enhetlig Access"


EdvinGroth:
In the compendium under 4.2.1 in the first question it asks you to connect the concept "Uniform Access"(number 14) to a description but in the rest of the compendium it is always referred to as "Enhetlig Access". The rest of concepts are in swedish except for this one, if this is not intended i would recommend changing the question and the answer sheet to "Enhetlig Access" instead of "Uniform Access".

bjornregnell:
Good catch! I aggree. It could even be "enghetlig acces (eng. uniform access)" to connect the terms. Do you want to make a PR yourself or do you want me to fix this?

bjornregnell:
If you run into troubles when fixing PR just ask a tutor in classroom when time permits.

bjornregnell:
fixed by #664

PR #664, opened 2022-09-20
Changed "uniform access" to "enhetlig access (eng. uniform access)"

EdvinGroth:


bjornregnell:
Have you built it locally and checked that the line does not overflow?

EdvinGroth:
I did not build it locally since i do not have a latex compiler. I tried doing it in overleaf but i reached the compile time limit so i couldnt check it there either since i also thought it might disrupt how the table looks. I will just remove the translation and try making a new pull request.

EdvinGroth:
It seems like it updates the pull request automatically so i won't make a new one, correct me if i'm wrong.

bjornregnell:
connected to #663

PR #674, opened 2022-09-26
fixed example 5.1.35 output

the0val:
I noticed the REPL output for Example 5.1.35 didn't match the actual output of running the code.

alt text


bjornregnell:
Good catch and nice fix!

Issue #678, opened 2022-09-29
[w05] confusion over not being able to access class methods through class name


snctfd:
I've noticed that students tend to make one common mistake in the exercises for w05; after implementing the `Pos` class, students tend to be "primed" to use the `Pos.moved` function when working on the implementation for `KeyControl.direction`, despite the fact that this function returns `(Int, Int)`. As a result, it's common for students to make the mistake of attempting to return e.g. `Pos.moved(-1, 0)` for the "left" case, likely because they don't actually have an instance of `Pos` to call `moved` on. This, of course, doesn't work. In previous years, I've also seen equivalent mistakes show up even later in the course.

What I would recommend to fix this is to add a small part in 6a) that encourages the student to try writing `Pos.moved(1,2)` in the REPL, and to reflect over why it doesn't work.

snctfd:
willing to provide PR if my suggested change seems good :)

Sabewey:
I have also seen several students trying to use `Pos.moved(...)` method to implement the `KeyControl.direction`. I would agree that perhaps a small exercise encouraging the students to try and make a method call to `Pos.moved()` without an instance of the class could help them question why this doesn't work and subsequently have a better understanding on how to approach the following question: `KeyControl.direction`. 

However, this might prompt the students to create a Pos instance in the `KeyControl` class instead...

bjornregnell:
Great! I agree that there is need for exercises that highlight how to use immutable and mutable structures here. Good if you can come up with a small but efficient exercise that tries to maximize understanding that does not deplete too much of the student's grit...  PR Welcome! @snctfd

snctfd:
PR created, feedback welcome! ^

bjornregnell:
merged! thnx!

PR #679, opened 2022-09-30
in exercises w05, encourage students to try Pos.moved and reflect over results

snctfd:
This is a first draft to solve #678. Feedback welcome!

I have not yet managed to compile the compendium to ensure that this looks good.

### DIFF
--- a/compendium/modules/w05-classes-exercise.tex
+++ b/compendium/modules/w05-classes-exercise.tex
@@ -296,6 +296,8 @@ scala> p.moved(0,1)
 val res1: blockbattle.Pos = Pos(1,3)
 \end{REPL}
 
+Testa även att anropa \code{moved} på klassnamnet, t.ex. \code{Pos.moved(0,1)}. Fungerar detta? Varför/varför inte? Reflektera över skillnaden mellan hur man anropar metoder i en klass jämfört med metoder i ett singelobjekt.
+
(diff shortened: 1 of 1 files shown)


bjornregnell:
I can merge and compile after review (currently you need old latex as installed in 20.04 or later or else allslides.tex fails). I will review this soon.

bjornregnell:
This looks great. It would be good if can add some text with explanation in the subtask solution as well?

snctfd:
> Classes don't have methods, only objects and class instances do, strictly speaking...

Good catch - I was a little uncertain as to what the specific terminology should be, I've changed it in accordance with your suggestion. 

I've also added an initial draft for an explanation to the solution. Remaining concerns:
* I feel like the wording of the explanation could be improved, but after writing out several variants this was about the best I could do. Thoughts, suggestions?
* As it is, I've put the explanation as a new SubtaskSolved, seeing as the existing TaskSolved for all of task 6 was just "Denna uppgift är laborationsförberedelse. Utvärdera dina lösningar genom egna tester i REPL.". Is this okay? Should I add some sort of context for what question my explanation is answering?
* As I understand it, you typically put errors in `\code` tags. I initially tried this with my text, but as the error message I listed is quite long and starts near the end of the line, this simply caused most of the text to disappear off the screen. For now, I instead put the error message in quotes, and simply `\code`-tagged `moved` and `Pos`. Is there a better solution?

bjornregnell:
LGTM

bjornregnell:
I will add "[till en referens] till klassen"

snctfd:
Oh yeah, that makes sense :)

bjornregnell:
https://github.com/lunduniversity/introprog/commit/603f70d24a0f3692c0a664e996626b66bdd2894f

bjornregnell:
many thanks for discovering and fixing this!

PR #680, opened 2022-09-30
corrected a typo

valterbergstrand:
added `s` to make a string into a string polarization on line 197


bjornregnell:
good catch!!

PR #695, opened 2022-10-12
Removed an unnecessary comparison

Filip-afKlinteberg:
Removed && i > 0 in the solution to assignment 7.2.5d as discussed in the discord-server

Since an arrays length can't be shorter than 0 the minimum value for result.length is 1.
if i < result.length then [...] i += 1 guarantees that i is at least equal to 1, so a comparison to see if it's larger than 0 is unnecessary.

The first commit is related to another issue, but I undid the change I made since that issue hasn't been solved yet.


bjornregnell:
Thanks! And good that you removed the unrelated change from this PR, but you are welcome with another PR on the `until` proposal..

Issue #692, opened 2022-10-11
Change the description of xs.indices in the quickref


Filip-afKlinteberg:
The current description of xs.indices in the quickref reads as follows: "Returns a Range extending from 0 to xs.length - 1".

During lectures and in the compendium (Assignment 7.2.3a) it's  instead described as a Range from 0 until xs.length. 

I think that it's better with a uniform description, so I suggest that the description in the quickref should be updated.

PR #696, opened 2022-10-12
updated description of xs.indices (Issue #692)

Filip-afKlinteberg:
Changed the description of xs.indices to match the description used in the compendium (Assignment 7.2.3a) and during lectures as approved in pull request #695.


bjornregnell:
Have you check that, when compiled with pdflatex, the line does not overflow?

Filip-afKlinteberg:
I compiled it in Overleaf and it doesn't overflow there.
Capture

PR #705, opened 2022-11-15
Update lect-w07-enums.tex

Hanyu-Lin:
Changed Console.BLUE to Console.BLACK to match enum case Black.


bjornregnell:
Well, is you have a black or dark background the blue is a better color to show a black playing card...  So this was actually deliberate. Or do you think it is too confusing?

Hanyu-Lin:
A student was confused by the difference, maybe you can add a comment to explain the thought process.

bjornregnell:
If there is room on the slide you can perhaps update your PR to keep BLUE but add a comment that "Blå färg syns på svart bakgrund" or similar -- IF there is room on the slide (?)...

Hanyu-Lin:
Updated my PR, but merge this minor clarification if you see it is necessary( and correct).

Issue #717, opened 2022-12-09
unimportant misspelling


isakstorme:
"litteral" and "literal" used inconsistently i believe

bjornregnell:
thanks for spotting this. Do you have any positional refs? (Or else a search will do when fixing this.)

bjornregnell:
In Swedish it should be with two t:s: https://sv.wiktionary.org/wiki/litteral

In English it should be with one t: https://en.wiktionary.org/wiki/literal

isakstorme:
Pages 26, 28, 38, 46, 49, 95 and 96 I think.

PR #723, opened 2023-01-03
fix 717 English literal used in Swedish sentences

fritjof-b:
`grep "literal" **/*.tex` in root-dir now only shows correct usages.


bjornregnell:
fixes #717

PR #730, opened 2023-06-15
update workflow file

fritjof-b:
- Improve readability
- Add functionality to manually trigger workflow
- Rename job
- Run on two latest Ubuntu LTS
- Boilerplate for uploading PDF artifacts generated by build

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,14 +1,48 @@
 name: Build
-on: [push]
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
 jobs:
-  Test:
-    runs-on: ubuntu-20.04
+  Compile-and-Build:
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, ubuntu-20.04]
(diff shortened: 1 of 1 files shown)


bjornregnell:
Many thanks!

bjornregnell:
You have question in the main.yml file by the end i an comment regarding the path. 

Where should it be uploaded? Is there a special artifact place under the repo created then on github or something? How are credentials or login handled or this is not needed?

fritjof-b:
No credentials needed. GH actions has its own store for artefacts (in this case PDFs). So when the build completes, you should be able to navigate to something like `Actions > Build > Artifacts > {foo.pdf, bar.pdf}`

So the idea was that if a build succeeds, we can view the generated compendium from that build. It's a minor thing, I was just unsure of where the PDFs are placed once generated since I was playing around in the Github environment on my fork... :-)

bjornregnell:
Ok great! Feel free to test with whatever path you think is necessary and let me know if I need to configure something as repo owner.

PR #747, opened 2023-09-05
Punkt tillagd mellan "xs" och "map" i Uppgift 5, W02

Aresiel:
Det saknas en punkt mellan `xs` och `main` här: (Uppgift 5, W02)
image
Med tanke på svarsalternativ antar jag punkten bör vara där.

### DIFF
--- a/quiz/QuizData.scala
+++ b/quiz/QuizData.scala
@@ -124,7 +124,7 @@ object QuizData {  // to generate tables for a concept connection quizes in late
       "\\code|ys.mkString              |" -> "ny sträng med alla element intill varandra",
       "\\code|ys.mkString(\",\")       |" -> "ny sträng med komma mellan elementen",
       "\\code|xs.map(_.toString)       |" -> "ny samling, elementen omgjorda till strängar",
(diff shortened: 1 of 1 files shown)


bjornregnell:
Have you tried it? At actually works due to infix operator notation which I briefly mentioned during this week's lecture. So I'm inclined to keep it. Or do you think it is better to wit to w03 to show infix notation? In that case also the space before the parenthesis should go away to be idiomatic...

Aresiel:
> Have you tried it? It actually works due to infix operator notation, which I briefly mentioned during this week's lecture. So I'm inclined to keep it. Or do you think it is better to wait until w03 to show infix notation? In that case also the space before the left parenthesis should go away to be idiomatic...
> 

You're right, I didn't think about that. It might be confusing to introduce it this early, especially since it looks like a typo, but there are also benefits of introducing it early. On the other hand it's also explicitly not recommended at the bottom of [this page](https://docs.scala-lang.org/style/method-invocation.html#higher-order-functions) on the Scala style guide, so it might be best to avoid due to that.

I'll make a commit removing the space after and you can decide whether to merge.

PR #753, opened 2023-09-19
Removed "new" keyword from w04-objects-exercise.tex

OlofBen:
New should not be used in Scala, except for in factory methods right?

### DIFF
--- a/compendium/modules/w04-objects-exercise.tex
+++ b/compendium/modules/w04-objects-exercise.tex
@@ -420,7 +420,7 @@ scala> // fortsätt så att en hel kvadrat ritas
 package hello
 
 object Main:
-  val w = new introprog.PixelWindow(400, 300, "HEJ")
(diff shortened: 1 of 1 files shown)


bjornregnell:
Well "should" is a bit strong. It's rather that you can skip it if you want.

bjornregnell:
I'm hesitating if we should keep new or not in this exercise as it makes the allocation explicit and it is a prequel to next week's theme on classes. What do you think? We could keep new here and add a comment or so that new can be skipped.

OlofBen:
You've got a point. However, the students have not seen the "new" keyword yet, so I think it can be a bit confusing for a new student.

bjornregnell:
Yes they have - there is a part during the lectures about classes as a prequel in week objects before next week on classes, and the see with and without new etc, and we talk about allocation on the heap and draw clouds etc.

bjornregnell:
Ok great. It is good if tutors explain what new does during exercises.

PR #764, opened 2023-10-29
fix lab code skeleton in w04

johannesnydahl:
The "s" is not needed here because string interpolation is not used.


bjornregnell:
Thanks for spotting this! I'd prefer if a string interpolator was used; hence that the data after + is inserted in the string in ${}. Would you like to fix that?

johannesnydahl:
Will do!

PR #778, opened 2023-11-16
Removed override modifer from exercise in W10

johannesnydahl:
image
The override modifier is not needed in this case because attributes x and y are not implemented in type Pair.


bjornregnell:
Well, it is still nice to use `override` so that you get help from the compiler to check if there actually *is* a thing called that to override...

Or do you have any specific reasons here not NOT use `override`, other than that it is unnecessary? @johannesnydahl

bjornregnell:
Perhaps if you could add a subtask then that asks if it works with override and why it is good to use override even if it is not necessary?
@johannesnydahl

johannesnydahl:
@bjornregnell That sounds like a great idea! Will update pull request.

bjornregnell (commented):
Should include sub-task on how override works.

PR #787, opened 2024-01-05
Add info to the quickref "snabbreferensen" that I believe belongs there

EliasAAradsson:


bjornregnell:
Thanks for suggestion! Have you checked that when pdf is built, the addition fits without overflowing to next age/column/outside page? 
(the quickref is quite crowded on limited pages by now and eventually we will not fit anything more if we don't throw out much of the less needed java-parts)

EliasAAradsson:
It does overflow at the moment. I don't see why there would be a problem if another page is added (then we could fit more interesting and useful stuff!). However, if adding another page is not an option, my suggestion is rewriting the entire section on Option, Some, None, and Try, Success, Failure. I believe this section can be shortened since it sort of says the same things about Options, Some, None, as it does about Try, Success, Failure. My suggestion would be to start the section with info that is the same for both Option and Try, then highlight differences. My contribution can be seen as one step in such a direction.

bjornregnell:
There is a hard limit on the number of pages to allow for booklet print. It is a big job to refactor the Java part to only fokus on the JDK. So in order to be able to merge this before a big re-write of the Java-part would like any addition also include reduction to fit in the addition. Taking on incremental changes that break the page limit is unfortunately not an option. Do you want to take a look at how to condense or make space for your addition or is there a solution to what you want to adress that takes less space?

EliasAAradsson:
Okay, then you can close this PR and I'll open an issue. Then hopefully I can do a rewrite of the part on Option, Some, None, and Try, Success, Failure, which will take equal or less space compared to now, while still providing space for my suggestion (and without sacrificing any of the already present stuff). It may take some time however before I can have a PR ready with this rewrite, since I have an upcoming exam to study for right now.

bjornregnell:
OK, closing. We continue discussion in issue.

PR #794, opened 2024-07-09
Add ??? to skeleton code in lab W04

johannesnydahl:
There is some skeleton code provided during the lab W04 that does not compile because some singleton objects are empty. My change adds the ??? placeholder to fix this and also briefly explains the question mark function.


bjornregnell:
:+1:

PR #800, opened 2024-08-24
Removed legacy boot suggestion, fixes #799

### LINKED ISSUE #799: Tools.md gives wrong advice on Legacy Boot
In the WINDOWS section, it says that when running Dual Boot it is required to turn on legacy boot in BIOS, which is incorrect because 1. Modern Linux distros run on GPT instead of MBR, and 2. It can cause the computer to stop booting the original OS.

I'd like to have it removed.

Annie-Wb:


bjornregnell:
Many thanks for spotting this! You can also come with another PR that adds your name to contributors.tex if you like (sorted by first name).

PR #806, opened 2024-09-11
Typo in compendium/modules/w02-programs-exercise.tex

LelleSwe:
rrayen -> arrayen


bjornregnell:
Thanks! Good catch. :)

PR #811, opened 2024-09-25
update lab task 5f in w01 to clarify type

johannesnydahl:
To generalize `kvadrat` and `stapel` so that the pen and fill color can be specified, one needs to know the type of color variables in Kojo. For students to figure the type out, they either have to study the Kojo source code or use Metals' type derivation. 

I think the type should be specified in this task to minimize confusion.

### DIFF
--- a/compendium/modules/w01-intro-lab.tex
+++ b/compendium/modules/w01-intro-lab.tex
@@ -320,7 +320,10 @@ stapel(42)
 
 \Subtask Skapa en abstraktion \code{rutnät} med lämpliga parametrar som gör att man kan rita rutnät med olika stora kvadrater och olika många kvadrater i både x- och y-led.
 
-\Subtask Generalisera dina abstraktioner \code{kvadrat} och \code{stapel} så att man kan påverka fyllfärgen och pennfärgen för kvadraterna som ritas ut. Färger i Kojo är av typen \code{Color}.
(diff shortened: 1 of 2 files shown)


bjornregnell:
Good point! Is it java.awt.Color or some other Kojo-package with own Color? I guess it would be good to state the full package path of the type if you want to take a look in the Kojo code and update your PR?

bjornregnell:
I've just checked. It is java.awt.Color. Probably we should import it in the file where other stuff from kojo is imported...

johannesnydahl:
Yes. Should I still change PR even if we add `import java.awt.Color` to workspace file?

bjornregnell:
Hmm. We have during the first week not mentioned import but the actual type is `java.awt.Color` so I suggest to update the PR with the last sentence being something like  
```
Färgerna i Kojo är av typen \code{java.awt.Color}. Typen är tillgänglig under namnet Color då namnet är importerat med \code{import java.awt.Color}.
```

johannesnydahl:
PR is updated.

bjornregnell:
I suggest that you change `importerat med export` till `gjorts direkt tillgängligt med export`

bjornregnell:
Och jag tror det fattas ett "i" före "vecka 4"

johannesnydahl:
My bad. Should probably be fixed now?

Issue #822, opened 2024-10-04
Many students disregard the alternative rectangle method in blockmole lab.


lego-eden:
In question **6c** of blockmole, the student is asked to write down the arguments to `fill( ... )` which would correspond to filling the same area of the window as the "loop version" I have encountered some students who have missed this question entirely. If it is considered very important for the learning goals of the lab then maybe it should be made clearer that even if you want to use the nested loop solution, you should also think through how the alternative could look.

The question could also be moved to the extra-question part of the lab if the focus should be on other parts of the lab. 

![image](https://github.com/user-attachments/assets/57e42911-5c29-47e6-a648-0752b47f06a0)

PR #823, opened 2024-10-04
move 6c from mandatory- to extra-questions

lego-eden:
This is in response to #822.


bjornregnell:
Thanks! I'm considering to wait with merging this until w06 is over so that we don't create a discrepancy with the print version and this will take effect for the next course instance.

lego-eden:
Sounds good!

PR #836, opened 2024-10-31
Make sure Strindberg doesn't come back and haunt us this Halloween

EliasAAradsson:


bjornregnell:
LOL. This is actually a funny typo :)

bjornregnell:
Perhaps I should let students vote on tomorrow's lecture if we should fix this or not :)

EliasAAradsson:
> Perhaps I should let students vote on tomorrow's lecture if we should fix this or not :)

Up to you :-)

PR #844, opened 2024-11-05
fix typo in w08 lab

johannesnydahl:


trilleplay:
https://svenska.se/tre/?sok=%C3%85nyo&pz=1 is a valid word

johannesnydahl:
Oops. Thanks for pointing out.

Issue #853, opened 2024-11-21
Remove inline comments in .gitignore files


RufusKoggRojder:
There are inline comments in a lot of the .gitignore files. This is not supported and breaks those ignores.

PR #854, opened 2024-11-27
fix gitignore inline comments

johannesnydahl:
See issue #853

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,10 +40,9 @@ about/**/.vscode
 **/target
 
 # Editor swap files
-
-## vim
+## Vim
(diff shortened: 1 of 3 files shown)


bjornregnell:
Many thanks! 

I was not aware that there is need for double `##`. Can you perhaps give me some link to where this requirement is documented, so I can read up on that before reviewing and merging this PR?

Also it seems that you have inconsistent use of spaces in comments and capital letters in comments (`##emacs` and `## Emacs` etc) so perhaps fix this?

johannesnydahl:
Will do

PR #858, opened 2024-12-11
fix typo

simonmalm:
fix typo in Projektuppgift: music (w13)


bjornregnell:
Thanks! Even two times the same mistake :blush:

PR #860, opened 2024-12-15
Add additional instructions for scaladoc

fgren:
Currently, the instructions in Appendix E doesn't mention that you need to have a local web server running (or use the patch script) to be able to view the generated documentation. The easiest way to do this is probably the Python http server, but since everyone doesn't have Python installed, a guide to using the patch script (slightly more complicated, but runs on Scala) is also included.


bjornregnell:
Thanks!

PR #868, opened 2025-06-13
fix minor typos and grammar

mnordahl:
Fixed a few typos and made some slight grammatical improvements in the first few pages of the compendium.
(Also added VS Code setting to disable LaTeX Workshop auto build, since it should be built via sbt)


bjornregnell:
cool! many thanks!

PR #889, opened 2025-08-31
Fix compilation error due to change in dependencies for kojo

EliasAAradsson:


bjornregnell:
partial fix: 
https://github.com/lunduniversity/introprog/commit/d4b52ad5734e5a341d12f0393821dd754cb02c09

bjornregnell:
https://github.com/litan/kojo-lib/pull/12

PR #893, opened 2025-09-05
Update Scala installation instructions for ARM devices

Annie-Wb:
Added instructions for ARM devices to install Scala tools.


Annie-Wb:
@bjornregnell to be fair it works even on x86 windows, since it's a java program, so if they run it, it will work the same as the other installer.

bjornregnell:
Aha!

PR #916, opened 2025-11-13
Use modern import syntax in the words lab

SLUCHABLUB:
Use `import nlp.*` instead of `import nlp._` since it is preferred in Scala 3.


bjornregnell:
Thanks!

PR #933, opened 2025-12-05
Give the music lab some love

SLUCHABLUB:
I've implemented some of the things I had in mind. Some changes may be more "controversial" than others so I'm prepared to cherry pick the commits.

TO-DO list:
- [x] Run `scala fmt`
- [x] Remove Scala 2 syntax
- [x] Use scientific pitch notation (middle C is C4 not C5)
- [x] Remove some unneeded brackets.
- [x] `import` stuff
 There are a lot of places where objects are fully qualified. I think it would be more readable if they were imported.
- [x] Use `math.floorMod` when calculating `relativePitchClasses`
- [x] Use more standard music terminology.
 For example change "relative pitch classes" to "simple intervals".
- [x] Rework some names
 Some names are (in my opinion) unnecessarily compressed. Whilst I can *accept* the use of single letter names (at least outside of public API), I would like to change `nbrOf*` and `to*Opt` to `numberOf*` and `to*Option` to align with the naming in `std`.
- [x] Use `enum`s in place of `Int`s
 There is a nauseatingly high usage of integers throughout the lab for things that should be `enum`s. On one hand it could prompt students to refactor the api themselves. On the other hand we should encourage the use of `enum`s and in my my experience most people are quite cautious when it comes to changing the given code.
- [x] Update the compendium
 The compendium needs to be updated to reflect (some of) the above changes.
- [x] Look over the `scalafmt` output and remove potential regressions in readability


bjornregnell:
Thanks! Do you think this is ready for review and potential merge now? @SLUCHABLUB

SLUCHABLUB:
It is ready for review.

bjornregnell:
Great! (I'll take a look after I'm ready with exam markings in ETSN15...)

Issue #942, opened 2026-07-12
autotranslate Option-D: ratification needed — MAYBE + PERSON clusters, and Färg→Suit/Colour context conflict


hmiddelk:
While clamping `slides/body/lect-w10-override.tex` for the English build
(Option-D `\ifswedish` clamps, see #940/#941), three items need BR ratification
before it can be clamped cleanly. Parking them here.

### 1. MAYBE cluster (an `Option`-analog demo)
`sealed trait Kanske[+A]` / `enum Kanske[+A]: case Någon(a: A); case Ingen`
- `Kanske` → `Maybe` (clear)
- `Någon` / `Ingen` → **naming decision needed.** Options:
  - `Just` / `Nothing` (Haskell-style) — but `Nothing` shadows `scala.Nothing`.
  - `Some` / `None` (Option-style) — familiar, but shadows stdlib (qualified as `Maybe.Some`/`Maybe.None`).
  - `Just` / `Empty` — avoids both stdlib collisions.

### 2. PERSON cluster
`case class Person(namn: String, ålder: Int)` and the `personExample*.scala`
examples (`Akademiker`, `Forskare`, `universitet`, `titel`, …).
- Proposed: `namn`→`name`, `ålder`→`age`, `universitet`→`university`,
  `Akademiker`→`Academic`, `Forskare`→`Researcher`, `titel`→`title`.
- Confirm names (esp. `Akademiker`/`Forskare`) before adding the cluster.

### 3. `Färg` context conflict (design)
The glossary maps `Färg`→`Suit` (cards, ratified for w06-patterns). But
`lect-w10-override` has `enum Färg { case Röd, Svart }`, which is a **colour**
context needing `Färg`→`Colour`, `Röd`→`Red`, `Svart`→`Black`. The
`apply-b0d` glossary is **global/token-exact**, so it can't render `Färg` as
both. The glossary already flags this: *"Färg→Suit is CARDS-CONTEXT-ONLY; a
kojo/colour file needs Färg→Colour (apply per-context)."*

Needs a decision on the mechanism, e.g.:
- per-file/per-context glossary override in `apply-b0d`, or
- a hand-written clamp for the colour `Färg` in `lect-w10-override` only, or
- rename the enum in the Swedish source to disambiguate.

Once 1–3 are settled, `lect-w10-override` can be clamped the same way as
`lect-w10-extends` (#940) and `lect-w06-matching` (#941).

bjornregnell:
[...]

PR #941, opened 2026-07-12
autotranslate: Option-D code clamps (lect-w06-matching)

hmiddelk:
Clamps Swedish code identifiers/strings to English in the pattern-matching
lecture, via `apply-b0d` `\ifswedish/\else` clamps.

> **Note — stacked on #940.** This branch builds on `option-d-clamps`, so until
> #940 merges the diff also shows its 6 commits. **The w06-specific changes are
> the top two commits:**
> - `8d021887` apply-b0d: w06-matching glossary cluster + `unapply` allowlist
> - `be60a15a` slides/lect-w06-matching: clamp code identifiers to English
> (Rebase/merge #940 first for a clean diff.)

## What's here (w06)
- New w06 glossary cluster (BR-ratified in session): identifiers (`smak`→`taste`,
  `lök`→`onion`, `livetsMening`/`LivetsMening`→`meaningOfLife`/`MeaningOfLife`,
  `mönster`/`resultat` 1..N, …) and the match result/prompt strings as **whole
  strings** (piecemeal would half-translate).
- `unapply` added to the allowlist (Scala extractor method — false positive).

## Result
- 24 code units clamped; fall-through 33→0.
- Slide-level gating verified: 27 slides → 13 fully-English, 2 fully-Swedish,
  **0 mixed**.
- Rebuilt `slides-en/lect-w06-en.pdf` model-free from the committed cache
  (0 model calls, cache-transparent): **3.3% Swedish**, code + result strings
  fully English, spacing correct.

PR #940, opened 2026-07-11
autotranslate: Option-D code clamps (lect-w10) + apply-b0d hardening

hmiddelk:
Renders Swedish code identifiers as English in the `\swedishfalse` build via
`\ifswedish/\else` clamps, and hardens the `apply-b0d` clamp tool.

## apply-b0d hardening
- Gate on code **identifiers only** — comments/strings are handled downstream by
  `translateCodeEnvBodies`, so they no longer false-skip a clamp.
- Keep string literals in the gate (avoids half-translated string data).
- **Slide-level all-or-nothing gating**: a slide clamps only if every code unit
  on it is clean — structurally prevents mixed-language slides.
- `--verify-examples`: render example `.scala` through the glossary and compile
  them — only compilable code passes.
- Fix `\fi` swallowing the following space (`\fi{}`), with `Latex.mask`
  absorbing the `{}` so clamps stay cache-transparent (round-trip verified).

## Content
- `slides/body/lect-w10-extends.tex`: 30 clamps → 9 fully-English, 3 fully-
  Swedish, 0 mixed slides. Verified by rebuilding `slides-en/lect-w10-en.pdf`
  model-free from the committed cache (8.8% Swedish, down from 43%).

New glossary clusters (w10 identifiers/strings) are flagged for ratification.

PR #943, opened 2026-07-17
autotranslate: translate \scalainputlisting example-file identifiers (shared CodeGlossary)

hmiddelk:
Translates Swedish identifiers in `\scalainputlisting`'d example programs (vego1-4.scala etc.) that leaked into the English compendium/slides chapter 10 (`object exempelVego1`, `trait Grönsak`, …). Single commit, rebased onto master.

## Root cause
Example programs are pulled via `\scalainputlisting`; the mirror translated them with `Code.translate` (comments + strings only, identifiers verbatim), and the inline `\ifswedish` clamps can't reach external `.scala` files.

## Change
- Promote the ratified apply-b0d glossary to a production module **`autotranslate/CodeGlossary.scala`** (single source of truth; apply-b0d aliases it).
- Mirror renames example `.scala`/`.java` identifiers via **`CodeGlossary.renderIds`** (token-exact identifiers only), then `Code.translate` handles comments + strings as before. `render` (id + str) is retained for the context-scoped slide clamps only.
- Add vego wrapper object names `exempelVego1..4 -> exampleVego1..4`.

## Verified
- `--verify-examples` COMPILE OK; mirror regenerates `compendium-en/examples/.../vego1.scala` as `object exampleVego1` / `trait Vegetable` / `class Cucumber(var weight)` / `isPeeled` — 0 model calls.

## Not covered (tracking #944)
`personExample*.scala` needs the PERSON cluster ratified (#942).

### DIFF
--- a/autotranslate/Code.scala
+++ b/autotranslate/Code.scala
@@ -65,13 +65,17 @@ object Code:
     * all, and stop the per-`--all` cache churn). So skip any line whose first non-space char is `%` — copied
     * verbatim, never translated. Default false: a real `.scala` line never starts with `%` (it's modulo, an
     * infix op: `a % b`), so the raw-source path (`--codetest` / workspace) is unaffected. */
-  def translate(src: String, tr: String => String, skipTexComments: Boolean = false): String =
+  def translate(src: String, tr: String => String, skipTexComments: Boolean = false,
+      alsoTranslate: String => Boolean = _ => false): String =
     val sb = StringBuilder(); val n = src.length; var i = 0
     // translate inner text of a comment OR string only if it looks Swedish — this skips already-English
     // library comments AND commented-out code (e.g. `//val w = new LifeWindow(20, 20)`), which must NOT
     // be sent to the model (wasteful + a source of masking-leak corruption).
+    // `alsoTranslate` opens a second gate for exact human knowledge: a no-åäö Swedish string the swedishish
+    // heuristic misses (e.g. "Skalas med skalare.") still routes to `tr` if it's a known Override key, so the
+    // Override resolves it. English strings aren't Override keys, so they still never reach the model.
     def piece(s: String): String =
-      if s.exists(_.isLetter) && swedishish(s) then tr(s) else s
+      if s.exists(_.isLetter) && (swedishish(s) || alsoTranslate(s.trim)) then tr(s) else s
     def texCommentLine(k: Int): Boolean = // first non-space char of the line at k is a LaTeX `%`
       var w = k
       while w < n && (src(w) == ' ' || src(w) == '\t') do w += 1
--- a/autotranslate/CodeGlossary.scala
+++ b/autotranslate/CodeGlossary.scala
@@ -1,0 +1,125 @@
+import scala.util.matching.Regex
+
+object CodeGlossary:
+  // VEGO cluster — ratified 2026-06-30 (vegomatch.scala + vego1-4 + w06-patterns), compile-verified.
+  // whole-identifier rename (token-exact). longest keys are fine; token replace is exact-match, not substring.
+  val id: Map[String, String] = Map(
+    "Grönsak" -> "Vegetable", "grönsak" -> "vegetable", "Grönsaker" -> "Vegetables", "grönsaker" -> "vegetables",
+    "Gurka" -> "Cucumber", "gurka" -> "cucumber", "ingenGurka" -> "noCucumber",
+    "Tomat" -> "Tomato", "tomat" -> "tomato",
+    "vikt" -> "weight", "ärRutten" -> "isRotten", "rutten" -> "rotten",
+    "slumpvikt" -> "randomWeight", "slumprutten" -> "randomRotten",
+    "slumpgurka" -> "randomCucumber", "slumptomat" -> "randomTomato", "slumpgrönsak" -> "randomVegetable",
+    "slumpbroccoli" -> "randomBroccoli", // exercise variant (w06): Broccoli stays (English), slump-prefix -> random
+    "Broccoli" -> "Broccoli", // identity: keep, but listed so a half-Swedish 'slumpbroccoli' can't slip the gate
+    "ärÄtvärd" -> "isWorthEating", "skörd" -> "harvest", "ätvärda" -> "worthEating", "ärÄtbar" -> "isEdible",
+    "anonymGrönsak" -> "anonymousVegetable",
+    "skalningsmetod" -> "peelingMethod", "skalfaktor" -> "peelFactor", "ärSkalad" -> "isPeeled", "skala" -> "peel",
+    // wrapper object names of the vego example programs (compendium/examples/.../vego1-4.scala).
+    "exempelVego1" -> "exampleVego1", "exempelVego2" -> "exampleVego2",
+    "exempelVego3" -> "exampleVego3", "exempelVego4" -> "exampleVego4",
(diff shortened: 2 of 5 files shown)


bjornregnell:
@hmiddelk Thanks! Nice work, and the root cause is right: the inline `\ifswedish` clamps cannot reach
external `.scala` files, so identifiers were always going to stay Swedish. Promoting the glossary to a real
module is the right call, and I like that apply-b0d now just aliases it. One source of truth.

I built it in a sandbox and ran your `CodeGlossary` over the corpus. It compiles clean and `--latextest` is
PASS with no failures, so nothing is broken. My comments are about what the mirror produces:

* The glossary is context-scoped, but the new call site applies it to the whole corpus:
  Your own note in `CodeGlossary.scala` says it: `Färg->Suit is CARDS-CONTEXT-ONLY; a kojo/colour file
  needs Färg->Colour (apply per-context)`. But `Main.scala` now runs `CodeGlossary.render` over *every*
  mirrored `.scala`/`.java`, and `mirrors = Seq("compendium" -> "compendium-en", "slides" -> "slides-en")`
  is the whole tree. So the map is applied per-corpus, not per-context.

  I ran `render` over every file the mirror would touch, and then checked what it does to `swedishish`,
  since `Code.translate` only sends a string to the model when `swedishish` is true, and `swedishish` leans
  first on the presence of åäö. Of the strings `render` changes, 45 go from swedishish true to false.

  Most of that is what you intend and it works nicely, since `render` already did the job:

  ```
  "gurka är gott!"             ->  "cucumber is tasty!"                ok
  "Skållas."                   ->  "Blanched."                         ok
  "Antal skördade grönsaker: " ->  "Number of harvested vegetables: "   ok
  ```

  The trouble is only where the glossary covers *part* of a string. Then `render` strips the åäöÅÄÖ that would
  have called the model, and the half-translated result ships:

  ```
  "Städa toaletter"  ->  "Clean toaletter"    (was "Clean restrooms" before this PR)
  "är akademiker"    ->  "is akademiker"
  ```

  So partial coverage is worse than no coverage: `render` is a partial translator that switches off full
  translation. Before this PR the `ä` brought in the model and it came out right; after, it can't. And the model-repair
  pass doesn't reliably save it: on a full build most of these ship as-is, and one (`JMutablePoint3D.java`)
  gets the mangled Swedish comment *translated into fluent, wrong English* ("no need for not this"), which is
  worse than a Swedish leftover because it reads like a real typo. `--verify-examples` passes on all of it,
  since it's inside string literals and still compiles.

  I think the fix is in your own title. It says *identifiers*, and the `id` map is token-exact and is
  the actual point of the PR, and it's fine. It's the `str` list that does the damage: bare `String.replace`
  over the whole file with keys as short as "hej", "gott", " är ", "inte ". ("hej" even fires inside "hejsan"
  in `hello-script.scala` and gives "hisan", which isn't a word in any language.) **What if the mirror path
  applies only `id`, and lets `Code.translate` keep doing strings and comments as it already does well?**
  Then "Städa toaletter" keeps its ä, reaches the model, and comes out right, and the vego identifiers still
  get renamed.

 **One thing I'd want *your view* on:** `"Skållas." -> "Blanched."` is in `str`, and that's real knowledge (skålla
  is blanch, not scald) that the model would get wrong. So `str` is doing override work. But we already have
  `Overrides.scala` for exactly that, and it doesn't have the substring problem. Maybe those entries belong
  there instead? WDYT?

* **Your vego1 "0 model calls" holds** and I reproduced it. One thing to flag though: a full `--all` build
  went from 9 calls on master to 22 with the PR, on the same cache. That fits the mechanism above: `render` rewrites the source text, which shifts the cache keys, so
  some strings that used to be cache hits now miss and hit the model. Not a blocker, just worth knowing the
  build is doing more work than the vego1 number suggests. (If we adopt the id-only fix above, this should
  mostly go away.)

* **Not your bug**, but I found it while testing: `--verify-examples` doesn't do what its name suggests when
  passed to the sbt run. It only exists in `apply-b0d.scala` and takes explicit file paths. If you pass it to
  `autotranslateProject/run` it silently falls through to a plain mirror run and verifies nothing. So "I ran
  --verify-examples" is only a real gate the way *you* invoke it (direct, with paths), not through the sbt task.
  I'll fix the tool to fail loudly on that; flagging so you know the gate is narrower than it looks. But I'll
  wait until this PR is merged, so it doesn't collide with your branch.

* **`// BR-ratified this session` is still there** (two places in `CodeGlossary.scala`). I think this is
  just the reword you already promised over on PR #941 not having made it across into this PR. Same fix:
  "ratified in session, pending BR confirmation" is perfect.

* **The `Trait betyder egenskap` override is mine, not yours. Pls remove it from this PR.** I looked,
  and our glossary has `Trait = Concept(sv = "trait", en = "trait", ...)`, so the term isn't translated in
  Swedish either. That slide line exists only to gloss the English word for Swedish students, so in English
  it has no job. "Trait means characteristic" isn't wrong exactly, it's just teaching an English reader a
  word they already have. I need to decide what that line should even be in the mirror (probably it should
  vanish, or become our glossary's own English line), and that's a pedagogy call I should make. You were
  right to mark it provisional. Sorry, this one is a gap in my own design, not a problem with your code.

Thanks again, this is a VERY valuable contribution! And the CodeGlossary module is a good piece of work.

This is my plan going forward:
* you fix stuff mentioned above that you agree with (or we discuss further here in this conversation)
* when this PR is merged I will:
  - fix --verify-examples, and
  - add the small check I wrote (the one that found the files above) to autotranslate/scratch/, so this is catchable next time

Let me know if you disagree with any of it, or just go fix'em :)

hmiddelk:
Both points addressed:

1. **Removed the `Trait betyder egenskap` override** — agreed, that's yours to make, and `Trait` is a Glossary concept term (handled via the authoritative tier), not an ad-hoc override.

2. **Fixed the corpus-wide `str` damage you flagged.** The mirror path now uses `CodeGlossary.renderIds` (token-exact **identifiers only**), not `render` (id + str). So `"Städa toaletter"` keeps its `å` and reaches the model, `"hej"` no longer fires inside `"hejsan"`, and JMutablePoint3D-style comments aren't mangled — while the vego identifiers still rename. `render` (id + str) stays only for the context-scoped slide clamps where the `str` list is ratified per file.

Rebased onto master (#940/#941 are in); #943 is now a single commit. `--verify-examples` still COMPILE OK.

hmiddelk:
Good catch — reworded all four `BR-ratified this session` comments in CodeGlossary.scala to `ratified in session, pending BR confirmation` (they rode along when I promoted the glossary out of apply-b0d; that's the reword I'd promised on #940/#941). The genuine `ratified 2026-06-30` VEGO/CARDS notes are left as-is.

bjornregnell:
Good! Merging now. I'll also purge the two stale Skållas cache rows post-merge

bjornregnell:
Follow-ups from the review are now on master:

  - --verify-examples passed to the sbt run now aborts loudly with the known-flag list, instead of silently doing a plain mirror run
  (any unknown flag aborts now).
  - The check that found the collateral during the review lives in autotranslate/scratch/glossary-blast-radius.scala. Default mode
  probes renderCodeIds; on the current corpus: 145 files scanned, 7 changed, all in ratified clusters, zero collateral.
  - Purged the two stale Skållas cache rows (the old model attempts were "Cheers." and "Scale carefully.", both wrong; your Overrides
  row is what ships now).

  Also noticed that renderCodeIds copies string literals verbatim, so the string-literal edge case from #944 is already handled. Nice work!

Issue #947, opened 2026-07-19
autotranslate: inline .tex code listings (compendium exercises) aren't translated by CodeGlossary


hmiddelk:
### Gap
Inline code in `.tex` — `\begin{Code}` blocks and `\code{...}` — gets English **only** via apply-b0d `\ifswedish` clamps. The mirror's `CodeGlossary.renderCodeIds` runs **only on external `\scalainputlisting`'d `.scala` files**, and `translateCodeEnvBodies` does comments/strings only (not identifiers). So a `.tex` file whose inline listings haven't been clamped shows Swedish identifiers in the English build even when `CodeGlossary` knows the translation.

### Concrete example (found in the rebuilt compendium-en)
`compendium/modules/w10-inheritance-exercise.tex`, the person-hierarchy task (~line 508):
```
class Person(val namn: String)
class Akademiker(namn: String, val universitet: String) extends Person(namn)
class Student(namn, universitet, program) extends Akademiker(namn, universitet)
class Forskare(namn, universitet, titel) extends Akademiker(namn, universitet)
```
`CodeGlossary` has `Akademiker→Academic`, `Forskare→Researcher`, `namn→name`, etc. (from #942), but this inline block isn't clamped, so it renders Swedish in chapter 10.

### Why a quick partial clamp is unsafe here
Compendium `.tex` has no `\begin{Slide}` grouping, so apply-b0d falls back to **per-unit** clamping — no all-or-nothing. Clamping the definition block alone would mix the exercise (definition `class Academic`, but the solution block `val akademiker = new Akademiker(...)` stays Swedish — lowercase `akademiker`/`forskare` are distinct tokens not in the cluster). This file also carries an unratified **ANIMAL cluster** (`djur`, `häst`, `läte`/`läten`, `väsnas`, `bondgård`, `räkna`, …) plus `IckeAkademiker` and a `"Doktorand"` string.

### Two ways forward
1. **Ratify + clamp**: ratify the person inflections (lowercase `akademiker`/`forskare`), `IckeAkademiker`, `"Doktorand"`, and an ANIMAL cluster; then apply-b0d-clamp `w10-inheritance-exercise.tex` (per-unit, reviewed for consistency). Same Option-D work, one big exercise at a time.
[...]

PR #945, opened 2026-07-19
autotranslate: MAYBE + PERSON glossary clusters (#942)

hmiddelk:
Adds the two BR-ratified clusters from #942 to `CodeGlossary`:
- **MAYBE** (lect-w10-override Option-analog): `Kanske→Maybe`, `Någon→Just`, `Ingen→Empty`.
- **PERSON** (personExample*.scala + lect-w10-override): `namn→name`, `ålder→age`, `universitet→university`, `titel→title`, `Akademiker→Academic`, `Forskare→Researcher`.

**Effect:** the `personExample*.scala` listings in compendium/slides chapter 10 now render English (via the mirror's `renderCodeIds`), completing the example-file work started for vego in #943. MAYBE is ready for the `lect-w10-override` clamp (next).

**Verified:** blast-radius check (`glossary-blast-radius … ids`) = 10 files changed, **0 collateral**; the person examples compile individually after render.

**Notes for BR:**
- `Färg→Colour` for the w10 colour enum is deliberately **not** added to the global glossary (stays `Färg→Suit`); it'll come via the per-file mechanism / hand-clamp (#942/#944).
- `trait Examinerad` (personExample3) isn't in the ratified list, so it stays Swedish — see #942.

PR #948, opened 2026-07-20
autotranslate: translate inline .tex code identifiers via CodeGlossary (#947)

hmiddelk:
The systemic fix for #947: the mirror now translates identifiers in **inline `.tex` code** (both `\begin{Code}` envs and inline `\code{}`), not just external `\scalainputlisting` `.scala` files — using the single-source-of-truth `CodeGlossary`.

> **Rebased onto #950** (branch `option-d-compile-gate`). #948 modifies `renderCodeIds`, and so does #950's s-interpolation fix, so #948 now sits on top of #950 to inherit that fix and avoid a merge collision. Until #950 merges, this PR's diff-vs-`master` includes #950's 3 commits (s-interp fix + compile gate + CI); **merge #950 first**, then #948 rebases down to its own commits (inline-code + ANIMAL cluster + the leak-check).

## Mechanism (Option-D via the mirror — English-only, Swedish source untouched)
- `translateCodeEnvBodies` runs `CodeGlossary.renderCodeIds` on **Scala-code env** bodies (`Code`/`CodeSmall`/`REPL*`/`lstlisting`/`Trace`/`Output`) — **not** `tikzpicture`/`pgf`/`forest`/`algorithm`/`exlatex`/`envi`/`comment`/`verbatim` (renaming node names / macros there would break refs; restricted via `scalaCodeEnvs`).
- Also renders inline `\code{}`/`\jcode{}`/`\lstinline{}` in prose, so a listing and its surrounding `\code` stay consistent.
- **Per-file overrides / opt-out** via `CodeGlossary.perFileId` / `optOut` (keyed on the mirror-relative path Main passes), so a context conflict (e.g. a colour `Färg` needing `Colour` not the global `Suit`) can deviate. Empty for now — the only source `enum Färg` are w06 cards (`Suit`, correct); the w10 colour enum is a `\code|...|` hand-clamp untouched by the brace-only inline pass.

## Verified
- `w10-inheritance-exercise` Task 5 now renders English: `class Person(val name)` / `class Academic(...) extends Person(name)` / `Student` / `Researcher`, and the inline `\code{Academic}`/`\code{Researcher}` in the subtask prose.
- Full `compendium-en.pdf` rebuilt (0 model calls): **4.6% Swedish** (was 4.8%), **0 LaTeX errors** — no tikz/algorithm breakage.
- `IckeAkademiker` stays Swedish (not ratified — flagged on #942).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,9 +3,24 @@ on:
   push:
     branches:
       - main
+  pull_request:
   workflow_dispatch:
 
 jobs:
+  # "Only compilable code should pass": every example .scala the mirror translates via CodeGlossary must
+  # still compile after the identifier rename (catches e.g. an s-interpolation left pointing at a renamed
+  # field). Runs from source (renders on the fly) — no mirror/texlive needed, so it's a fast, standalone gate.
+  Example-compile-gate:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      - name: Setup scala-cli
+        uses: VirtusLab/scala-cli-setup@v1
+      - name: Compile gate for mirror example translation
+        run: scala-cli run autotranslate/scratch/verify-mirror-examples.scala -- .
+        shell: bash
+
   Compile-and-Build:
     strategy:
       matrix:
--- a/autotranslate/CodeGlossary.scala
+++ b/autotranslate/CodeGlossary.scala
@@ -111,10 +111,34 @@ object CodeGlossary:
   def renderCodeIds(s: String): String =
     val out = new StringBuilder; val code = new StringBuilder; var i = 0; val n = s.length
     def flush(): Unit = if code.nonEmpty then { out ++= renameAll(code.toString); code.clear() }
-    def copyDelim(q: Char): Unit = // copy a "..." or '...' literal verbatim (handles \-escapes)
+    def copyChar(): Unit = // copy a '...' char literal verbatim (handles \-escapes); never interpolated
       out += s(i); i += 1
-      while i < n && s(i) != q do { if s(i)=='\\' && i+1 < n then { out += s(i); i += 1 }; if i < n then { out += s(i); i += 1 } }
+      while i < n && s(i) != '\'' do { if s(i)=='\\' && i+1 < n then { out += s(i); i += 1 }; if i < n then { out += s(i); i += 1 } }
       if i < n then { out += s(i); i += 1 }
+    // an interpolator prefix (s"…"/f"…"/raw"…") is any identifier char immediately before the opening quote.
+    def isInterp: Boolean = out.nonEmpty && (out.last.isLetterOrDigit || out.last == '_')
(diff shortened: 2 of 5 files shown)


hmiddelk:
**Heads-up — Option-D status, w10 fixes, and one source bug**

A few things landed / need your eyes:

### Rebases (stack refreshed after #945 merged)
- **#946** (`lect-w10-override`) rebased onto `master` — now independent, just its own two commits.
- **This PR (#948)** rebased **onto #950**, not `master`, because both edit `renderCodeIds` (the s-interp fix vs. #948's per-file param). Basing #948 on #950 inherits the s-interp fix and avoids a merge collision. **Merge order: #950 → #948.** Until #950 lands, this PR's diff shows #950's three commits too.

### ANIMAL cluster (w10 Task 3 `trait Djur`) — added here, per your OK
- Identifiers: `Djur/Ko/Gris/Häst/väsnas/skapaDjur/bondgård → Animal/Cow/Pig/Horse/makeNoise/createAnimal/farm`.
- Sounds: `Muuuuuuu/Nöffnöff/Gnääääägg → Mooooo/Oink/Neigh`. These are string *data*, which `renderCodeIds` deliberately left verbatim, and inline `.tex` never runs through `Code.translate`/Overrides — so they had **no** translation route. Added a ratified `codeStr` map (whole-literal exact match, never substring) applied inside string literals.
- **Enforcement:** the mirror gate now also fails if a ratified `codeStr` key survives untranslated in any `.tex` Scala-code region (293 `.tex`, 0 leaks) — so a sound can't silently regress, and a new sound not ratified into `codeStr` is caught.

### #949 (concept-list typos)
Extended beyond topptyp/bottentyp: also `inmixning supertyp → mixin supertype` (same half-translation failure class). Title/body updated. Still your call to ratify wording.

### New issue #952 — a source bug, needs your decision
The `inmixning supertyp` concept item is actually two concepts fused by a **missing comma** in `plan/Plan.scala:134` (the W10 "Innehåll" string, split on commas to build the list). It corrupts the Swedish chaphead list *and* both plan tables (the EN plan table even keeps it Swedish). One-char source fix proposed in the issue. **Note:** #949's `inmixning supertyp` override is a band-aid on this — once the comma is fixed, it and the purged cache row become removable. Tell me whether to drop them now or after the source fix. I didn't touch the Swedish source myself since it's content structure.

bjornregnell:
The comma is fixed at the source (53730d5f) and regenerated, so yes: please drop the inmixning supertyp override and its cache purge from Overrides, the two concepts now translate on their own.

bjornregnell:
One thing to double check after the ANIMAL cluster: väsnas and Djur also occur in other tasks of w10-inheritance-exercise (the Fyle bird tasks around line 1012 and the Djur/Katt/Hund block), where Fyle, läte, Katt and Hund are not in the glossary. 
Those listings now render mixed, e.g. def makeNoise with print(läte * 2), and differ from the cache which translated väsnas as behaves there. Could you eyeball those renderings and either extend the cluster or scope it via perFileId?

PR #956, opened 2026-07-21
Option-D w10 follow-ups: scope ANIMAL, ratified terms, protect \ifswedish clamps

hmiddelk:
Post-merge follow-ups to #948/#949 from BR's review, all verified in a rebuilt `compendium-en.pdf`.

## Changes
1. **Drop the `inmixning supertyp` override** — BR fixed the missing comma at source (`plan/Plan.scala`, 53730d5, closes #952), so the concept-list item is now two concepts (`inmixning`→mixin, `supertyp`→supertype) translating on their own. The #949 workaround is removed. (Left the wrong `inmixning supterp` cache row purged, not restored — it's dead now; say if you'd rather have a literal revert.)

2. **Scope the ANIMAL cluster to `perFileId["w10-inheritance-exercise"]`** (was global). `Djur` also occurs in `lect-w11-context`'s generics demo (`class Katt/Hund extends Djur`), where a global `Djur→Animal` produced mixed `class Katt extends Animal`. Scoping keeps w11 fully Swedish (deferred) while Task 3 still translates. The rest of the cluster occurs only in this file.

3. **Mirror leaves code inside `\ifswedish` clamps untranslated** — `translateCodeEnvBodies` was running `renderCodeIds` on code inside hand-written `\ifswedish…\else…\fi` clamps, mutating the Swedish branch into a mixed listing (the hand-clamped Fyle example rendered `def makeNoise … print(läte*2)`). New `Latex.ifswedishRanges` (nesting-aware) gives the clamp ranges; both the env pass and inline pass skip matches inside them.

4. **Ratified terms** (BR): `IckeAkademiker→NonAcademic`; `Doktorand→"PhD student"` (title string → `codeStr`); lowercase `akademiker→academic`, `forskare→researcher` (Task 5 REPL val names).

5. **Clamp the Fyle `räknaLäte` solution** — the one Fyle block that wasn't `\ifswedish`-clamped, so `väsnas` collided (must be `makeNoise` in Task 3 but `holler` in Fyle — same file, unresolvable by glossary). Hand-clamped using the Fyle example's established terms (`väsnas→holler`, `läte→call`, `Fyle→Fowlk`) plus new `räknaLäte→callCount`, `antalLäten→numberOfCalls`. Re-termed BR's existing Pjodd `\else` clamp (was `callTally`/`callCount`) to the same scheme so both solutions read consistently. Swedish branches kept byte-identical.

## Verification
[...]

Issue #951, opened 2026-07-20
autotranslate: extend the compile gate to inline .tex code (not just \scalainputlisting .scala)


hmiddelk:
Follow-up to the compile gate added in #950.

### What's covered now
`autotranslate/scratch/verify-mirror-examples.scala` (CI job `Example-compile-gate`) renders every **`\scalainputlisting`'d example `.scala/.java`** through `CodeGlossary.renderCodeIds` and compiles it standalone, failing on any translation-introduced regression. This covers the reported break (personExample1/2 = compendium-en 10.1.27/10.1.29 — the s-interpolation bug).

### The gap
It does **not** compile-check **inline `.tex` code** — the `\begin{Code}` / `\code{}` blocks translated by the mirror pass in #948. Those aren't standalone compilable units: a listing is often a fragment (a single `def`, a REPL transcript with `scala>`/output/error lines, a snippet that references types defined in a neighbouring block). So the same class of bug (e.g. an s-interpolation or a rename that doesn't compile) could slip through in an inline listing.

### Options
1. **Extract + assemble**: pull each Scala-code env body, filter out REPL prompt/output/error lines, group blocks that share a slide/exercise into one compilation unit, and compile. Fiddly (REPL transcripts, fragments, cross-block refs) and inherently heuristic.
2. **Round-trippable listings only**: compile only the `\begin{Code}` blocks that look like complete top-level definitions (skip `REPL*`/`Trace`/`Output` transcripts and bare fragments). Narrower but reliable.
3. **Author-marked**: a marker on listings that are meant to compile, gated; everything else skipped.

Lower priority than the `.scala` gate (most identifier-bearing code lives in the `\scalainputlisting` example files, which #950 covers). Filing so it's tracked. WDYT on the approach?

bjornregnell:
I did some counting to ground this. Across compendium and slides there are about 1800 Scala code envs, but only about 230 of them (13 percent) are
  actually rewritten by the CodeGlossary pass, and that is the only population the gate needs to protect. Of those 230, 140 are REPL transcripts,
  65 look like complete units and 25 are fragments.

  I think option 2 is best, but:

  1. Scope it to envs where renderCodeIds actually changed something, not all inline code.
  2. Use the same regression semantics as your #950 gate: compile the Swedish body and the rendered English body, skip if both fail, flag only when
  the Swedish compiles but the English does not. In my sampling about a third of complete-looking bodies do not compile standalone (neighbour
[...]

PR #957, opened 2026-07-22
autotranslate: phase-2 REPL compile gate (#951)

hmiddelk:
Phase 2 of the inline `.tex` compile gate (#951), as BR outlined — REPL transcripts.

> **Stacked on #956** (phase-1 gate). Until #956 merges this PR's diff-vs-`master` includes #956's commits; **merge #956 first**, then this rebases down to its single commit.

## What it does
For each rewritten, non-`\ifswedish` `REPL`/`REPLnonum`/`REPLsmall` transcript:
- **Split on `scala>` prompts** — prompt-stripped input plus its indented continuation lines; output lines (result echoes `val x: T = …`, `resN`, error `|` / `-- Error`, blanks) are dropped.
- Concatenate the transcript's inputs into one `object` body and compile the **Swedish** vs **rendered-English** programs under the phase-1 regression rule: regression = Swedish compiles but English doesn't; skip if both fail.
- Clamp-aware (`Latex.ifswedishRanges`), changed-only, uses the mirror's per-file overrides.

**Safety:** line classification is *identical* for the Swedish and English transcripts (parallel structure), so a mis-split can only cause a **skip**, never a false regression.

## Tested (both paths)
- Injected `Tomat→Tomato` duplicate-class transcript → caught, **exit 1**.
- Self-contained `Gurka→Cucumber` transcript compiles → **non-vacuous**.
- `\ifswedish`-clamped copy of the same break → **skipped**.

## Clean run (real repo)
`phase 2 = 12 ok, 85 skipped, 0 regressions` (85 skipped are mostly output-only blocks with no `scala>` input, or transcripts that reference defs from a neighbouring Code env). The 12 OK include `w10-inheritance-exercise.tex:261` — the Task 3 farm REPL — so the ANIMAL rename is now compile-checked. `-- . --list` prints the per-transcript OK/skip breakdown.

Remaining known limitation: transcripts that depend on a preceding Code env compile in neither language and self-skip; wiring that cross-env context is a possible phase 3 but wasn't in scope.

### DIFF
--- a/autotranslate/CodeGlossary.scala
+++ b/autotranslate/CodeGlossary.scala
@@ -49,17 +49,15 @@ object CodeGlossary:
     "Kanske" -> "Maybe", "Någon" -> "Just", "Ingen" -> "Empty",
     // PERSON cluster (personExample*.scala + lect-w10-override) — BR-ratified 2026-07-19 (#942).
     "namn" -> "name", "ålder" -> "age", "universitet" -> "university", "titel" -> "title",
-    "Akademiker" -> "Academic", "Forskare" -> "Researcher",
+    "Akademiker" -> "Academic", "Forskare" -> "Researcher", "IckeAkademiker" -> "NonAcademic",
     // one-off example identifiers surfaced by the #944 coverage sweep — BR-ratified 2026-07-19 (#942).
     "Examinerad" -> "Graduated",                                       // trait Graduated { val title: String }
     "kastaTärningTillsAllaUtfallUtomEtt" -> "rollDieUntilAllOutcomesExceptOne",
     "BaklängesHandler" -> "BackwardsHandler",
-    // ANIMAL cluster (w10-inheritance-exercise Task 3, `trait Djur`) — BR-agreed 2026-07-20.
-    // Inline .tex code (REPL/Code envs + prose \code{}) → translated by the mirror's inline pass (#948).
-    // The onomatopoeia string literals (Muuuuuuu/Nöffnöff/Gnääääägg) are English sounds too — via `codeStr`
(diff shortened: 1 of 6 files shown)


hmiddelk:
@bjornregnell ready for review — phase 2 of the #951 gate (REPL transcripts).

**Merge order:** stacked on #956, so **#956 first**, then this rebases to its single commit.

How it works: each rewritten, non-`\ifswedish` `REPL*` transcript is split on `scala>` prompts (prompt-stripped input + indented continuations; output lines — `val x: T =…`/`resN`/`|`/`-- Error`/blanks — dropped), its inputs concatenated into an `object` body, and the Swedish vs rendered-English programs compiled under the same regression rule as phase 1 (regression = SV compiles, EN doesn't; skip if both fail).

The design point that makes this safe: SV/EN line-classification is **symmetric** (parallel transcript structure), so a mis-split can only cause a skip — never a false regression. Borne out on real data.

Tested both paths: injected `Tomat→Tomato` duplicate-class transcript caught (exit 1); self-contained `Gurka→Cucumber` transcript compiles (non-vacuous); `\ifswedish` copy skipped.

Clean run: **phase 2 = 12 ok, 85 skipped, 0 regressions** (85 = output-only blocks with no `scala>` input, or transcripts referencing a neighbouring Code env). The 12 OK include the Task 3 farm REPL, so the ANIMAL rename is now compile-checked. `-- . --list` prints the per-transcript breakdown.

One note: transcripts that depend on a preceding Code env self-skip (no cross-env context) — that'd be a phase 3 if you think it's worth it; I kept this PR to what you scoped.

Issue #958, opened 2026-07-23
review comments on PRs 956 and 957


bjornregnell:
@hmiddelk

Thanks for two excellent PRs! I have just merged 956 and will soon merge 957 after 956 has been greended by gh actions. The Example-compile-gate is a required status check on master, so the gates now actually protect the compendium as you intended.

Did an agent-supported review of the code after merging. Everything below is in order of priority. Only item 1 has correctness weight, the rest are small things to take at leisure, or to skip if you disagree.

1. **REPL classifier is ASCII-only, which can break the SV/EN symmetry** (`autotranslate/scratch/verify-mirror-examples.scala:147`). The result-echo pattern `(val|var)\s+\w+:.*=.*` uses `\w`, which does not match Swedish letters. So a typed val/var line like `var räknaLäte: Int = 0` is not classified as output on the Swedish side, while its renamed twin `var callCount: Int = 0` is on the English side. That breaks the invariant that line classification is identical for both transcripts, and in the worst case a mis-split drops a definition from one program only, giving a false regression or masking a real one as a skip. Today it is latent (the one identifier pair of that shape is clamp-protected), but it is a one-line fix: make the regex Unicode-aware, e.g. prefix with `(?U)`.

2. **Phase-1 regression entries lack line numbers** (same file, around line 115). The ok and skip lists print `file:line`, but a regression prints only `file (env)`. The regression list is the one a maintainer must act on, so it should be the easiest to locate. Phase-2 already does this right.

3. **Warn on an unbalanced `\ifswedish` clamp** (`autotranslate/Latex.scala:148-155`). `ifswedishRanges` scans raw text, so a literal `\ifswedish` in a comment or verbatim block opens a bogus range that can swallow everything to end of file, and following envs silently go untranslated and ungated. The failure direction is conservative (skip, never mistranslate), but a stderr warning when a range ends at end-of-file would make it visible.

4. **Name the tuple in `ifswedishRanges`** (`Latex.scala:148`). The return type `Seq[(Int, Int)]` has two same-typed slots, so the compiler cannot catch a swapped start/end. A named tuple `Seq[(start: Int, end: Int)]` self-documents and call sites read better. Latex.scala is shared production code, so it is worth it there.

5. **Usage header is stale** (`verify-mirror-examples.scala:29-34`). It still says only `-- <introprog-root>`, but the root is now optional (defaults to `.`) and `--list` exists.

[...]

PR #959, opened 2026-07-24
autotranslate: address review #958 (gate correctness + polish)

hmiddelk:
Addresses the agent-supported review in #958. Off `master` (both gate PRs merged). Thanks for the thorough review — and for enabling branch protection so `Example-compile-gate` is a required check.

**1 — correctness (the only one with weight).** The REPL result-echo classifier `(val|var)\s+\w+:.*=.*` was ASCII-only, so `var räknaLäte: Int = 0` (SV) wasn't classified as output while `var callCount: Int = 0` (EN) was — breaking the SV/EN classification symmetry that guarantees a mis-split can only skip, never false-flag. Fixed with a `(?U)` prefix. Verified: the SV line now classifies as output (was `false`), and a REPL fixture carrying that echo line self-compiles (`ok`) instead of self-skipping on a duplicate `var`.

**2** — phase-1 regression entries now print `file:line (env)` (like phase-2), so the actionable list is easy to locate.

**3** — `Latex.ifswedishRanges` warns on stderr when a range runs to EOF (unbalanced `\ifswedish`, e.g. a stray one in a comment/verbatim) — already conservative (skip, never mistranslate), now visible.

**4** — `ifswedishRanges` returns a named tuple `Seq[(start: Int, end: Int)]`; call sites (Translate + the gate) read `r.start`/`r.end`.

**5** — refreshed the gate usage header (root optional, `--list` documented).

**6** — noted the per-file scaffold repetition across the three loops is deliberate for a scratch tool (a shared HOF would add more indirection than it saves).

**7** — left as-is per your note (Code/CodeSmall don't take optional args).

Main project + gate both compile; negative REPL test still catches an injected regression (exit 1). Running the full clean gate locally to confirm 0 regressions with the new classifier (CI will also run the required check).

### DIFF
--- a/autotranslate/Latex.scala
+++ b/autotranslate/Latex.scala
@@ -145,13 +145,28 @@ object Latex:
     * The mirror's code-env / inline-code translation uses this to LEAVE hand-clamped code alone: both branches
     * of a `\ifswedish<sv>\else<en>\fi` clamp are already final, so translating inside one would mutate the
     * Swedish branch (e.g. rename `väsnas`->`makeNoise` while its neighbours stay Swedish -> a mixed listing). */
-  def ifswedishRanges(s: String): Seq[(Int, Int)] =
-    val out = scala.collection.mutable.ArrayBuffer[(Int, Int)]()
+  def ifswedishRanges(s: String): Seq[(start: Int, end: Int)] =
+    val out = scala.collection.mutable.ArrayBuffer[(start: Int, end: Int)]()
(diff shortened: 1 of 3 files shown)


bjornregnell:
All seven points addressed, and the Unicode fix is exactly right: verified that the classifier now treats a Swedish-typed echo line as output on both sides, so the SV/EN symmetry holds and the negative test still catches an injected regression. Nice that it recovered two transcripts' coverage as a side effect. Merging. Thanks for the careful pass.

Issue #960, opened 2026-07-24
compendium-en: Swedish leaks — ratify w10 glossary terms; decide on prose type-names and stale cached Swedish prose


hmiddelk:
Reviewing the English `compendium-en.pdf` (and the w10 slides) surfaced Swedish leaks that fall into three distinct classes. Class 1 is fixed and just needs ratification; classes 2 and 3 need a decision.

## 1. Code identifiers missing from the glossary — FIXED, please ratify
Inline `\code{}` placeholder type-names in `lect-w10-extends` that weren't in `CodeGlossary`. Added (scoped via `perFileId` to `lect-w10-extends`, since `Typ` is a common token that would bleed into e.g. `lect-wjava`'s `\code{Klassnamn<Typ>}`):

| Swedish | English | where |
|---|---|---|
| `EnVissTyp` | `SomeType` | 10.1.28 `isInstanceOf[..]`/`asInstanceOf[..]` |
| `Typ` | `Type` | 10.1.30 `isInstanceOf[Typ]` |
| `KlassensNamn` | `ClassName` | 10.1.26 single-public-class file name |
| `bastypensNamn` | `baseTypeName` | 10.1.26 lowercase-initial multi-content file name |

→ PR to follow. **Please ratify these four English terms** (or suggest alternatives).

## 2. Type names in prose / `\texttt{}` — DECISION NEEDED
The mirror only translates `\code{}`/`\jcode{}`/`\lstinline{}`. Type names written another way stay Swedish:
- `lect-w10-extends:298` — `\textit{\textbf{\texttt{Grönsak}}}` (a `\texttt`, not `\code`) → renders "Grönsak" though `Grönsak→Vegetable` is glossary'd.
- `lect-w10-extends:934` — plain prose "…tilldelning Student, men också **Akademiker** och Person" (no code macro) → the model keeps "Akademiker".

Options: (a) convert these `\texttt{TypeName}`/bare type names to `\code{TypeName}` (semantically correct, and then the glossary translates them — but it changes the **Swedish** build's markup); (b) `\ifswedish`-clamp each (English-only, Swedish byte-identical, per-occurrence); (c) leave. This is a Swedish-content call, so it's yours.

## 3. Stale cached Swedish-fallback prose — DECISION/PLAN NEEDED
[...]

PR #961, opened 2026-07-24
compendium-en: fix missing ToC + w10 placeholder type-names

hmiddelk:
Two independent compendium-en output fixes found while reviewing the English build. See #960 for the ratification ask and the prose/cache decisions.

## 1. Missing table of contents (build fix)
`compendium-en.pdf` had no ToC. `runPdfLatexCmd` ran pdflatex once; a ToC needs ≥2 passes (pass 1 writes `.toc`, pass 2 typesets it, a 3rd may be needed since inserting the ToC shifts `\pageref` page numbers). The Swedish tasks survive on one pass because `compendium/`/`slides/` keep `.aux`/`.toc` across builds, but the English mirror dirs are regenerated fresh, so the single pass wrote `.toc` and never read it.

Fix: `runPdfLatexCmd` gains `maxPasses` and loops **until LaTeX stops asking to "Rerun"** (a tiny latexmk), capped — the minimum passes to converge, not a guessed 2 or 3. Swedish tasks keep cap 1 (unchanged); English tasks use cap 4. Verified on a fresh-dir rebuild: the ToC renders.

## 2. w10-extends placeholder type-names (pending ratification — #960)
Inline `\code{}` placeholders in `lect-w10-extends` missing from the glossary, scoped via `perFileId` (not global, so `Typ` can't bleed into `lect-wjava`'s `\code{Klassnamn<Typ>}`):

| Swedish | English | where |
|---|---|---|
| `EnVissTyp` | `SomeType` | 10.1.28 |
| `Typ` | `Type` | 10.1.30 |
| `KlassensNamn` | `ClassName` | 10.1.26 |
| `bastypensNamn` | `baseTypeName` | 10.1.26 |

Verified via `renderCodeIds`: all four translate under the `lect-w10-extends` path; `Typ` stays Swedish under `lect-wjava` (scoping holds). **English terms pending your ratification (#960).**

Model-free, English-only (Swedish build byte-identical). No new gate regressions.

### DIFF
--- a/autotranslate/CodeGlossary.scala
+++ b/autotranslate/CodeGlossary.scala
@@ -203,6 +203,16 @@ object CodeGlossary:
       "Djur" -> "Animal", "Ko" -> "Cow", "Gris" -> "Pig", "Häst" -> "Horse",
       "väsnas" -> "makeNoise", "skapaDjur" -> "createAnimal", "bondgård" -> "farm",
     ),
+    // lect-w10-extends type-casting section (10.1.28/10.1.30): placeholder type-names in the inline
+    // \code{isInstanceOf[..]} / \code{asInstanceOf[..]} examples. Scoped here (NOT global) because `Typ` is a
+    // common token — a global Typ->Type would bleed into e.g. lect-wjava's \code{Klassnamn<Typ>} (mixed).
+    // BR-ratified 2026-07-24: EnVissTyp ("en viss typ") -> SomeType; Typ -> Type.
+    "lect-w10-extends" -> Map(
+      "EnVissTyp" -> "SomeType", "Typ" -> "Type",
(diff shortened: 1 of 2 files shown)


bjornregnell:
The rerun-until-converged loop is the right fix for the missing ToC, and keeping the Swedish tasks at cap 1 (byte-identical) while only the English mirrors loop is exactly the scoping I'd want. Terms ratified in #960. One tiny thing before I merge: the comment at CodeGlossary.scala:209 says "BR-ratified 2026-07-24" while the PR still says pending — now that I've ratified, that's fine, just confirm the date. Merging.

(Minor, no action needed now: the rerun check keys off "Rerun to get" / "Label(s) may have changed" but not longtable's "Table widths have changed. Rerun LaTeX." — worst case is the old single-pass behaviour, so no regression; worth a note only if a wide table ever shows stale widths.)

PR #962, opened 2026-07-25
autotranslate/scratch: purge helper for stale Swedish cache fallbacks (#960)

hmiddelk:
The tool I offered on #960 for your class 3+4 model run — targeted re-translation of the stale Swedish fallbacks without a full `--clean`.

## What it does
Removes **stale `key==value` Swedish fallback rows** (units cached as their own Swedish when no model was reachable — they render Swedish forever because it's a cache *hit*). Dry-run reports **699** in `translate-cache.tsv` + **16** in `translate-code-cache.tsv`. Purging turns them into cache *misses* so a model run re-translates them.

## Usage (atomic, model-enabled)
```
scala-cli run autotranslate/scratch/purge-stale-swedish-cache.scala -- .            # dry run (report)
scala-cli run autotranslate/scratch/purge-stale-swedish-cache.scala -- . --write    # apply
sbt "autotranslateProject/run --all"     # WITH Ollama/modly reachable -> re-translates the purged units
git diff autotranslate/translate-*cache.tsv   # review the new English, then commit
```
(Alternative: `run --all --clean` re-translates all ~15k rows; this purge does just the ~715 stale ones.)

## Safety
- **Read-only without `--write`** — single guarded `os.write.over`; verified sha256 unchanged across a dry run.
- Only `key == value` **and** Swedish (å/ä/ö or a whole-word Swedish marker) rows are removed; real translations are never touched.
- **Header warns loudly**: never commit a purged-but-not-retranslated cache — with no model, the purged units fall back to Swedish (PDF unchanged) but live fallbacks jump ~9 → ~715 and **break the cache-only CI prose-leaks ratchet** (baseline = the live-measured 9). Purge only as step 1 of the atomic workflow above.

No production code touched — a scratch tool for your model run. Happy to fold in `--retry-fallbacks` semantics instead if you'd prefer the mirror own this.

### DIFF
--- a/autotranslate/scratch/purge-stale-swedish-cache.scala
+++ b/autotranslate/scratch/purge-stale-swedish-cache.scala
@@ -1,0 +1,67 @@
+//> using scala 3.8.4
+//> using jvm 21
+//> using dep com.lihaoyi::os-lib:0.11.8
+
+// Purge STALE Swedish->Swedish fallback rows from the mirror's translation caches, so a subsequent
+// MODEL-ENABLED run re-translates them into real English. A fallback row has an identical Swedish key and
+// value (key == value) — it was "translated" when no model backend was reachable and then cached as its own
+// Swedish, so it renders Swedish forever (a cache HIT on the frozen Swedish), even in a model-enabled build.
+// Purging it turns the unit back into a cache MISS, which the model then translates. (See issue #960, classes 3+4.)
+//
+// !!! DO NOT COMMIT A PURGED CACHE WITHOUT A PAIRED MODEL RUN !!!
+// Purge + re-translate is ATOMIC and must happen on a MODEL-ENABLED machine. The purged units become cache
+// MISSES; with no model reachable a miss falls back to Swedish, so:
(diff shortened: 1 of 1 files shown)


hmiddelk:
Closing in favour of your `SwedishScore` (now in master) — the right foundation.

Your review showed my ad-hoc detector is unsafe for a precision-first purge:
- the raw å/ä/ö rule flags the **CC BY-SA licence block** (an accented name in the copyright line) as Swedish, so `--write` would delete a good row;
- the 32-word regex misses ~24 diacritic-free Swedish prose rows your scorer catches.

And your conceptual correction is the key one: the model-gate (recall-first — a miss is a permanent leak) and the purge (precision-first — a false positive deletes a good row) want **opposite thresholds**, so the shared thing is the **word lists + evidence**, with each caller thresholding the **3-level verdict** — not one boolean, which is what I'd assumed.

So: we're aligned on the cache mechanics (stale `key==value` hits render Swedish; purge → miss → re-translate, atomic with a model), but the detector is better as yours. When you do the class 3+4 model run, the purge should be rebuilt on `SwedishScore` at the Swedish-only (precision) threshold. Thanks for building the scorer and the careful diff.

PR #963, opened 2026-08-11
autotranslate: phase-3 cross-env session gate (#951)

hmiddelk:
@bjornregnell phase 3 of the #951 gate — the cross-env context we parked behind the `@later` note on #957.

For the REPL class this is your own suggestion carried one step further: phase 2 wraps a transcript's inputs in `object ReplWrap:`; phase 3 wraps *the preceding envs plus the body* in `object Session:`, under the same #950 regression rule.

## What it does

A recovery pass over the 113 phase-1/2 skips. Those bodies compile in neither language not because they are broken, but because they reference names defined in a **neighbouring env**.

- **Session** — from the nearest structural boundary (`\Task` / `\WHAT` / sectioning / `\begin{Slide}` / `\QUESTBEGIN` / `\*Tasks`) to the target. **Coarser than the parked sketch, deliberately:** `\Subtask` and `\SOLUTION` are *not* boundaries, because a solution's code routinely references the hierarchy printed in its own task statement — w10 Task 5's `val person = new Person(...)` needs the `class Person` env from before `\SOLUTION`.
- **Window, not whole preamble** — the k preceding units for the **smallest k whose Swedish program compiles** (cap 4). This is the other departure from the sketch and it is forced: successive subtask solutions *redefine* the same types, so prepending the whole session stacks three `Person` definitions into an ambiguity error. k is chosen on the source side alone, so unit selection stays identical for SV and EN — a mis-grouped boundary can only skip, never false-flag.
- **Clamps** — a whole `\ifswedish…\else…\fi` collapses to one context unit, its `\ifswedish` branch feeding the Swedish compile and its `\else` the English one (new `Latex.ifswedishBranches`). That pairing is what catches a hand-clamped *definition* and an auto-translated *reference* that disagree on a name.

## Coverage

Gated bodies **37 → 52 (+41 %)**. Phase 3: **15 recovered**, 50 offered no in-session definition, 33 have no code at all, 15 unresolved within a 4-env window, **0 regressions**. Your step-0 estimate was ~20 recoverable; 15 landed.

Phases 1–2 reproduce byte-identically, which is how I know they are untouched:
example **10 ok / 0 reg** · leak-check **293 / 0** · phase-1 **23 ok / 30 skip / 97 deferred / 0 reg** · phase-2 **14 ok / 83 skip / 0 reg**.

## Non-vacuous *and* effective

Per the pattern from #956/#957 — a clean run proves nothing by itself.

Injected `perFileId["lect-w11-context"] = Map("compare" -> "xs")`, i.e. renamed `Ordering.compare` in one slides file. That break is **invisible to all three existing gates** (example, phase-1 and phase-2 all stayed green) but fatal once the session is assembled: phase 3 flagged both `lect-w11-context.tex:898 (Code)` and `:905 (REPL)`, and the gate exited 1. Injection reverted before committing.

That specimen also illustrates the mechanism nicely — the REPL at `:884` defines `case class Gurka` and `xs`, the `Code` at `:898` defines the `given Ordering[Gurka]`, and `:905` is a bare `xs.sorted`, recovered at k=2.

## Two supporting changes in the same commit

Both were needed to make the above workable, so I would rather flag them than bury them.

1. **`compiles()` now uses the persistent bloop daemon** instead of `--server=false`. Measured **2.2 s warm vs 13.5 s cold** per snippet; the whole gate runs in **5 m 16 s** instead of well over an hour — so this also makes the CI job cheaper than it is today. Isolation is unchanged: still a fresh temp directory per snippet, only the daemon is reused. Because a wedged daemon would fail a compile for reasons unrelated to translation, and a failed English compile is exactly what this gate reads as a build-breaking REGRESSION, every would-be regression is re-confirmed with a cold isolated compile before the build is failed. **This touches phases 1–2's machinery**, not just phase 3's — verdicts are identical (numbers above), but it is a shared code path and you should know that.
2. **`--report <file>`** streams the report to a file as the run proceeds: useful as a CI artifact, and necessary when the runner captures stdout rather than echoing it.

## Caveats

- **`Latex.ifswedishBranches` has no unit test** — only transitive coverage through the gate.
- **The 33 "no code at all"** are genuinely output-only transcripts: nothing to compile, out of scope rather than uncovered.
- **The 50 "no context offered"** are the external class you predicted (~24 measured at step 0); they stay covered by the reserved opt-out marker if one ever needs it.
- **Compile ≠ correct**, as before — this proves the English builds, not that the terms are right.
- Phase 2's transcript reader is **untouched**; phase 3 has its own, because Scala 3's `|` continuation prompt is currently classified as output and truncates a multi-line entry to nothing. Honest measurement: that fix moved exactly **one** transcript from "no code" to "has code" and gated **zero** extra bodies. Kept because the defect is real and phase 3's context units depend on the reader, but it did not pay off on this corpus, and I over-sold it before measuring.

## AI assistance

Per the README policy. I used Claude Opus 5 on this change: it drafted the session/window algorithm and the `ifswedishBranches` helper, and drove the verification loop. I set the scope (phase 3 alone, then the transcript reader as a separate decision), and I checked myself that the phase-1/2 baselines reproduce exactly and that the injected break really does fail the build. Two impact claims the agent made were wrong and are corrected above: the transcript-reader payoff, and an initial cost estimate for phase 3 that turned out to be an artefact of the cold compiler path rather than a property of the new phase.

### DIFF
--- a/autotranslate/Latex.scala
+++ b/autotranslate/Latex.scala
@@ -169,6 +169,33 @@ object Latex:
         out += ((start = k, end = e)); i = e
     out.toSeq
 
+  /** Every `\ifswedish…\fi` block split into its two BRANCHES: `elseAt` is the offset of the block's
+    * TOP-LEVEL `\else` (nesting-aware — an `\else` belonging to a nested `\if…` does not count), or
+    * `end` when the block has none (a Swedish-ONLY clamp). So `[start, elseAt)` is the Swedish branch
+    * and `[elseAt, end)` the English one.
+    *
+    * Used by the phase-3 cross-env gate to pick the right side of a HAND-CLAMPED context env: a
+    * definition is often clamped (`\ifswedish class Ko … \else class Cow … \fi`) while the REPL that
+    * REFERENCES it is auto-translated, so the Swedish compile needs the `\ifswedish` branch and the
+    * English compile the `\else` branch. Classification is on the raw .tex, so it is identical for
+    * both languages. */
+  def ifswedishBranches(s: String): Seq[(start: Int, elseAt: Int, end: Int)] =
+    ifswedishRanges(s).map: r =>
+      var i = r.start + "\\ifswedish".length
+      var depth = 1
+      var el = -1
+      while i < r.end && el < 0 && depth > 0 do
+        if s(i) == '\\' && i + 1 < r.end && isCmdLetter(s(i + 1)) then
+          var j = i + 1
+          while j < r.end && isCmdLetter(s(j)) do j += 1
+          val nm = s.substring(i + 1, j)
+          if nm == "fi" then depth -= 1
+          else if nm == "else" && depth == 1 then el = i
+          else if nm.startsWith("if") && nm != "iff" then depth += 1
+          i = j
+        else i += 1
+      (start = r.start, elseAt = if el < 0 then r.end else el, end = r.end)
+
   /** Mask the source. Returns (maskedText, spans, itemIdx) where itemIdx is the set of placeholder
     * indices that are `\item` markers (used to split bullet lists into per-item translation units).
     * With stripEng=true, `\Eng{...}` is removed. */
--- a/autotranslate/scratch/verify-mirror-examples.scala
+++ b/autotranslate/scratch/verify-mirror-examples.scala
@@ -24,26 +24,59 @@
 // envs (Code/CodeSmall/lstlisting) the mirror rewrites in .tex — skipping \ifswedish-clamped code. And a
 // PHASE-2 REPL GATE: REPL* transcripts split on `scala>` prompts, inputs concatenated into an object and
 // compiled under the same regression rule (SV/EN classification is symmetric, so a mis-split only skips).
-// See those sections below.
 //
-//   scala-cli run autotranslate/scratch/verify-mirror-examples.scala -- [introprog-root] [--list]
+// Plus a PHASE-3 CROSS-ENV SESSION GATE (#951): a recovery pass over the phase-1/2 skips, which are mostly
+// bodies that reference names defined in a NEIGHBOURING env. Each is retried with the smallest window of
+// preceding same-session envs whose SWEDISH program compiles — the window is chosen on the source side, so it
+// is identical for both languages — and a hand-clamped context env contributes its \ifswedish branch to the
+// Swedish compile and its \else branch to the English one. That pairing is what catches a clamped DEFINITION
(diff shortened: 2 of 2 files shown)


bjornregnell:
Thanks Hans, this is a careful piece of work. The negative test is what makes it
  reviewable: renaming `Ordering.compare` in one file, showing all three existing
  gates stay green, and showing phase 3 flags both the Code env and the dependent
  REPL, tells me far more than a green run would. Reproducing phases 1-2
  byte-identically to prove the shared `compiles()` change is behaviour-preserving
  is the right way to land a change that touches a shared path.

  Two comments, neither blocking.

  1. `Latex.ifswedishBranches` and the missing unit test. You flagged this yourself,
     and I agree it should have one, for a specific reason: the nesting logic keys
     off `nm.startsWith("if") && nm != "iff"`, which treats any command beginning
     with `if` as opening a conditional that a `\fi` will close. That holds for
     `\ifkompendium`, `\ifx`, `\iftrue`. It does not hold for `\ifthenelse`, which
     takes brace arguments and no `\fi` -- one of those inside a clamp would leave
     `depth` permanently raised and the top-level `\else` would never be found, so
     the block would be classified as Swedish-only and the English branch silently
     dropped. I checked: there is no `\ifthenelse`, `\ifdefined`, `\ifnum` or
     `\ifcase` anywhere in `compendium/` or `slides/` today, so this is latent
     rather than live. A test pinning the three cases (no `\else`, nested `\if`
     with its own `\else`, and a non-`\fi` `\if`-prefixed command) would fix the
     shape of the contract before someone adds one.

  2. The cold re-confirm is one-directional. Re-confirming a would-be regression
     with an isolated compile means a wedged daemon cannot produce a false FAILURE,
     which is the right priority. But nothing re-confirms a warm PASS, so daemon
     state that made a snippet compile when it should not have would produce a
     false GREEN -- and for a gate, a missed break is the more costly direction.
     Have you considered cold-verifying a small random sample of passes, or the
     whole set on a nightly/scheduled run? Cheap insurance for the one case the
     current design does not cover.

  Also worth knowing, since it lands next to your clamp work: I have started using
  `\ifswedish ... \else ... \fi` inside section headings on the Swedish side
  (`\section{\ifswedish Valfri tentamen \else Optional written exam \fi}`). Those
  are exactly the hand-clamped/auto-translated pairing your phase 3 targets, so
  your gate and my heading clamps are going to meet.

I'm currently working on the English-side build after some Swedish-side additions,
  and I plan to merge this once that work has landed.

PR #971, opened 2026-09-02
Updated open folder guide and added workspace

Annie-Wb:
Just before your last commit @bjornregnell I started working on improving the kojo description, so I merged the two changes.

It now specifically explains to make a new folder, open it and OBS to not open just the file or Downloads.

I also added downloading workspace.zip as part of it, since it already has kojo. Wondering if we should just skip the kojo part and use directly workspace.


Annie-Wb:
Also considering we can add c3pu in the install list, including how to run it, what do you think?

PR #973, opened 2026-09-05
Update macOS tool installation

### LINKED ISSUE #972: update macOS install instructions
The macOS install instructions are a bit outdated (and suspicious):

* Ghostty might be to prefer over iTerm2 (opinionated, ready to hear critiques)
* there is no clear reason to use bash over default Zsh for macOS, since the incompatibilities will be of no concern for people who are taking their first course in programming

This has already been discussed with Björn, so I will drop a PR despite no visible discussion being present.

christofergartner:
* Remove any mentions of opting into bash for mac.
* Suggest Ghostty as first choice of replacement for Terminal on mac. It is generally a faster and more elegant tool for the needs of an introprogger.

closes #972

### DIFF
--- a/web/tools/tools.md
+++ b/web/tools/tools.md
@@ -225,14 +225,20 @@ Verifiera att det blev installerat:
 brew --version
 ```
 
-#### **Valfritt**: iTerm2
-MacOS kommer med en inbyggd terminal, men många föredrar att använda iTerm2 som är mer flexibel.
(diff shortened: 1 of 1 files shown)


bjornregnell:
Thanks! I think it would be better to just have one new terminal, so I guess ghostty and skip iterm? And also it would be helpful to explain *how* to open it not just say "open it"... Would you like to fix that? @christofergartner

Issue #968, opened 2026-08-30
lect-w11-context.tex: code comment refers to non-existent object MinKontext


hmiddelk:
`slides/body/lect-w11-context.tex:713` has a code comment that refers to an object which does not exist:

```scala
scala> import EnAnnanNamnrymd.given    //importerar alla givna värden i MinKontext
```

The object actually being imported is defined 15 lines earlier in the same file, at line 698:

```scala
object EnAnnanNamnrymd:
  given String = "tagen för givet" //namn behövs ej, det räcker med typ
```

**Version:** line 713 in release `v2026.6`, and unchanged on `master` at `a924f330` (2026-08-25).

**Why this looks like a leftover rather than a spelling mistake:** `MinKontext` occurs exactly once in the whole repository — in this comment — and is defined nowhere under `slides/`, `compendium/` or `workspace/`. It is also not a misspelling of the real name, since "min kontext" and "en annan namnrymd" are different phrases. It reads like a remnant of an earlier version of the example in which the object had a different name.

**Suggested fix:**

```scala
scala> import EnAnnanNamnrymd.given    //importerar alla givna värden i EnAnnanNamnrymd
```

**Question:** is renaming the comment the fix you want, or did this example once have a separate `MinKontext` object that should be reintroduced to make the import-scope point clearer? I am happy to submit the one-line pull request if the rename is what you prefer.

To be clear about scope: the Swedish itself is not the problem here. Per the contribution guide, Swedish identifiers in lectures and exercises are deliberate. The comment simply names an object that exists in neither language. This surfaced while clamping the w11 identifiers for the autotranslated English mirror; the mirror deliberately reproduces the same dangling reference (translated as `MyContext`) so that the Swedish and English sides stay faithful to each other and the defect stays visible in both, rather than being silently corrected on the English side only.

*Assistant note, per the "Making contributions" guidelines in the README:* I used Claude Code to sweep this lecture for Swedish identifiers remaining in the English mirror, which is how the comment surfaced. I checked the following myself: that `MinKontext` appears nowhere else in the repository, that the object at line 698 is `EnAnnanNamnrymd`, and that the line is identical in `v2026.6` and current `master`.

bjornregnell:
Good catch! I'll look into this after the semester start rush is over. Should be fixed on the Swedish side. PR welcome.

PR #976, opened 2026-09-17
autotranslate: clamp the w11 context lecture example identifiers

hmiddelk:
`lect-w11-context` was the last of the *clamped* lectures still reading Swedish in the English mirror — w06, w10 and the `\scalainputlisting` example files were done earlier, and this completes that series. It clamps 28 scoped identifiers, 3 promoted to the global map, 5 code strings and 4 code comments, with each case routed through the mechanism it actually resolves in.

To be clear about what it does **not** finish: lectures Option-D never covered still contain Swedish identifiers. w03 is the clearest case — `lect-w03-functions` and `w03-functions-exercise` have no `\ifswedish` clamps at all, so `öka`, `förnamn`/`efternamn`, `hemlis`/`gissning`, `hälsa`/`fleraAnrop` and others render Swedish in the English compendium — both the lecture slides and the chapter-3 task REPLs, and 3.1.35 comes out mixed, with English strings around Swedish identifiers. An earlier revision of this description claimed w11 was the last lecture *containing* Swedish; that was wrong, and the corrected scope is as stated here.

## Why four mechanisms rather than one

Picking the wrong one is a **silent no-op**: the entry looks like a fix, changes nothing, and nothing warns. I hit that twice while building this, which is also why your new override-orphans gate is such a good addition.

| Text lives in | Mechanism | Why not the others |
|---|---|---|
| identifiers in code | `CodeGlossary.id` / `perFileId` | token-exact, code regions only |
| whole quoted literals | `codeStr` | exact full content; an unquoted REPL echo needs an `id` entry instead |
| code comments | `Overrides` | `str` is deliberately kept out of `renderCodeIds`, and inline `.tex` comments resolve through the code-cache, which Overrides beats |

## Scoped vs global

Most of the cluster is scoped to this file, because `Bur`, `Ros` and `behandla` are short or common tokens and `lect-w06-equals` still has unclamped `case class Hund/Katt` that must stay uniformly Swedish.

`Toalett`/`Upptagen`/`Ledig` went global instead: the covariant-enum demo has no context conflict, the three tokens occur nowhere else in `slides/` or `compendium/`, and `namnrymd -> namespace` style terminology is authoritative in `glossary/concepts.scala` anyway.

Renaming matches whole identifier **tokens**, never substrings, so a compound is a different token from the word inside it — `Robothund` does not pick up `Hund -> Dog`, so it has its own `RobotDog` entry, as does `kattBur -> catCage`. The same exactness is what keeps `Ros -> Rose` off `Rosenqvist` and `"hej" -> "hi"` out of `hejsan`.

## Two naming judgements worth your eye

- **`framkalla` -> `elicit`, not `summon`.** The following slide introduces Predef's `summon` as *"a generic variant of the method `framkalla`"*; clamping to `summon` would make that sentence tautological and put two `summon`s in one demo.
- **`tulipanaros` deliberately left Swedish.** It is the portmanteau the slide links out to (`sv.wikipedia.org/wiki/Tulipanaros`) — the joke itself rather than an example name. `Tulpan`/`Ros` are clamped around it. Happy to clamp it too if you would rather the English deck drop the reference.

Also fixed: a leaked mask placeholder, `... :(<-C0>`, that rendered literally in the English w11 deck. It was the only such leak in the repo.

## The MinKontext comment is mirrored, not corrected

#968: that comment names an object existing nowhere in the material (the real one is `EnAnnanNamnrymd`). The override translates the dangling reference rather than fixing it, so both language sides stay faithful and the defect stays visible in both. **The moment you fix the Swedish, this entry becomes an orphan and should be deleted** — I will happily send that as the #968 PR you invited, dropping this entry in the same change.

## Verification

On this base, a full `autotranslateProject/run --cache-only`:

```
done. model calls: 0, fallbacks: 9, overrides: 485, cache: 15483
[overrides] OK: measured 45 within baseline 45
[cache-only] OK: measured 9 within baseline 9
```

So no new orphan debt — all four override keys match live units — and the fallback floor is untouched. Every clamped identifier, string literal and comment was then checked line by line in the regenerated `slides-en/body/lect-w11-context-en.tex`.

**What this does NOT fix**, so remaining Swedish in the lecture does not read as an oversight:

- 4 prose `\item`s, and two REPL block rows whose cached values the guards reject for mangled identifiers (`EnNamnrymd` -> `EnNamespace`, and `ejscala>` rewritten to `scala>`, which destroys the marker for deliberately-broken code). Model work, not clamp work.
- Line 191, the `\else` slide variant `\code|enum Toalett[+T] { ... }|`. `Translate.inlineCodeRe` is brace-only, so pipe-delimited inline code is never clamped. Only the deck is affected; the compendium uses the `\ifkompendium` branch. I left it alone because widening that regex changes behaviour for every `\code|...|` in the corpus — your call whether that is worth a follow-up.

## A question about direction

You mentioned on #969 that you intend to use clamped `\ifswedish` as much as possible for future Swedish-side changes. Since this is the last of the lecture-by-lecture clamp work, it seems the right moment to ask: **do you want this glossary-driven style continued for new material, or should new demos be hand-clamped with `\ifswedish` instead?**

They pull in different directions — `\ifswedish` keeps both languages in one source and needs no glossary entry, while the glossary approach keeps the Swedish source untouched and renames only the mirror, which scales across files but needs a ratified entry per name. I am happy either way; I would just rather not add glossary entries you would prefer not to maintain.

## Assistant note

*(per "Making contributions" in the README, rather than a commit trailer)* I used Claude Code for this work: it swept the lecture for Swedish identifiers, proposed the entries, and diagnosed which mechanism each case needed. I checked the results myself against built output rather than the diff — the numbers above come from the unscoped run's own report, and I read the regenerated mirror line by line for every entry. The two silent no-ops mentioned above were caught that way, not by the tooling.


hmiddelk:
Correction to my own description, which I have just edited.

The opening line claimed that `lect-w11-context` was *the last lecture whose Scala demos still read Swedish in the English mirror*. That is wrong. It was the last of the **clamped** lectures; it is not the last lecture containing Swedish.

I found this by reading the built English compendium rather than the diff. w03 is untouched by Option-D: `lect-w03-functions.tex` and `w03-functions-exercise.tex` contain **no `\ifswedish` clamps at all**, and none of their identifiers are in the glossary, so they render Swedish in the English build:

- **3.1.6** (named arguments) — the REPL declares `förnamn`, `efternamn`, `förnamnFörst`, `ledtext`.
- **3.1.12** (local functions) — `skrivLedtrådOmEjRätt`, `gissning`, `hemlis`, `inteRätt`.
- **3.1.35** (loop with recursion) — the worst of the three to read, because it is *mixed*: the slide title and the interpolated strings came out English while the identifiers stayed Swedish, and two strings are still `är för stort` / `är för litet`.
- **Chapter 3 tasks** — the REPL sessions declare `öka`, `skrivNamn(efternamn, förnamn)`, `hälsa`, `fleraAnrop(antal, f)`.

Across the whole built PDF there are 112 REPL lines carrying Swedish diacritics, and that is a lower bound: diacritic-free identifiers such as `hemlis` and `gissning` are not counted by that scan.

None of this changes what this PR does or how it was verified — the gate numbers stand, and w03 was never in scope here. Only the claim about scope was wrong, and I would rather correct it before you read it than have you discover the overstatement yourself.

Separately, while checking chapter 6 I found a real mistranslation that is not in this PR: in `lect-w06-matching` and `lect-w07-seq`, `\Emph{gard}` renders as **garden** where the term is a pattern **guard** — and the Swedish source even supplies the English itself as `\Emph{gard} \Eng{guard}`. It only breaks where the word sits alone inside `\Emph{}`, i.e. the isolated-emphasis child unit, exactly the "poisoned child" pattern from your info-week00 entries; everywhere else in those two files it is correctly "guard". A single child-level override fixes both occurrences. I will send it as its own PR once this one is merged, since both touch `Overrides.scala` — unless you would rather I fold it in here.

bjornregnell:
Thanks, this is careful work and the annotations are what make it reviewable.

  I checked the claims the entries rest on and they hold. Toalett, Upptagen and Ledig occur only in lect-w11-context.tex and nowhere
  in compendium, so global is safe here. MinKontext really is a dangling name: one occurrence at :713, and the import on that same
  line is EnAnnanNamnrymd.given. And lect-w06-equals does still carry unclamped case class Hund and Katt, so scoping the animal
  cluster rather than promoting it is right.

  The framkalla to elicit argument convinces me. Keeping the pun rather than pre-empting the next slide's summon is the pedagogically
  correct call, and I would not have wanted summon there.

  One thing to settle before this lands. The PR handles two source defects two different ways. MinKontext is mirrored on purpose so
  the defect stays visible on both sides, which you argue well. But vattning is silently corrected to watery in the English only, so
  the Swedish keeps an error the English has already fixed. I would rather we were consistent. Since vattning appears twice, at :433
  and :457, it is a settled spelling rather than a slip, so please leave the English faithful and I will fix the Swedish source
  instead. Same for MinKontext, which I will correct in the slide.

  On tulipanaros: leave it unclamped. The link to the Swedish article is the joke, and a translated version of it would not be.

* I will soon report measured result in the next comment*

  Before I merge I need the full-run number. The four new Overrides keys each risk being an orphan, and override-orphans-baseline.txt
  is a ratchet at 45 that only allows a fall. Could you run a full gen and report the orphan count, confirming it is still 45 or
  lower? A scoped run will not measure it.

bjornregnell:
I measured the ratchet myself so you do not have to: the gate passes. A full cache-only run on your branch merged with master
  reports overrides OK, measured 45 within baseline 45, and cache-only OK at 9 within 9. All four new keys match, confirmed in the
  generated mirror rather than assumed: the ÄrDennaTyp comment at line 51, the covariant Bur comment at 301, the KAOS line at 305 and
  the MinKontext import at 713. The key count goes 392 to 396 and the orphan count does not move, which only works if none of the
  four is an orphan.

  Worth knowing that your reasoning about the durable home is confirmed: the old leaky row is still in translate-code-cache.tsv at
  line 538, placeholder and all, and the output carries your override text instead. The override beats the cache exactly as you
  argued.

  One thing I did not measure: the heading-map floor. The pdftk on my box is the snap build and could not read the sandbox, so gen
  failed closed rather than writing a map. Nothing in this PR is heading-related so I am not worried, but I would rather say it is
  unmeasured than imply it passed.

  One interaction to handle before this lands, and it is my change causing it, not yours. If I fix vattning to vattnig in the source,
  your codeStr key stops matching, silently, and the English deck goes back to showing the Swedish word. Nothing gates that: the
  ratchet counts Overrides keys only, so a dead CodeGlossary key has no check at all. The same goes for MinKontext if I correct that
  slide. So let us sequence it: keep your entries as they are, I will hold both source fixes until after this merges, and then change
  the source and the glossary key in one commit. If you think a CodeGlossary orphan check is worth having as well, say so and I will
  open an issue for it.

bjornregnell:
Answers to your three questions, and one thing I have just caused that you need to know about.

  First the thing I caused. I have pushed 09f82404 on master, fixing the typo vattning to vattnig in lect-w11-context.tex at lines
  433 and 457. Tomat has syrlig, an adjective, so Gurka should have vattnig rather than the noun vattning. That means your codeStr
  key "vattning" no longer matches anything. Once you pick up master it will silently do nothing and the English deck will show the
  Swedish word again. Please change the key to "vattnig" in this PR. I am sorry for the ordering, I fixed the source before thinking
  it through.

  Worth saying that nothing gates this. The orphan ratchet counts Overrides keys only, so a dead CodeGlossary key has no check at
  all. If you think a CodeGlossary orphan check is worth having, say so and I will open an issue. I measured the ratchet on your
  branch merged with master before that commit and it passes, 45 within 45 with all four override keys matching, and the vattnig
  change does not affect that number.

  On direction, question one. Prefer clamped \ifswedish for new material. Both languages stay in one source, there is no glossary
  entry to ratify, and the entry cannot silently stop matching the way we have just seen. So please treat this PR as the last of the
  glossary-driven batches. It should still merge, since it clamps material that already exists and it is verified.

  On gard, question two. The source is correct already, so nothing needs fixing in the .tex. Both occurrences read \Emph{gard}
  \Eng{guard}, at lect-w06-matching.tex:247 and lect-w07-seq.tex:350, and there are none in the compendium.

  I would rather not take the two override entries though. Look at what the mirror does today in lect-w06-matching-en.tex. Line 246
  says "Matching with guard" and line 267 says "with guard", both correct, and line 247 says \Emph{garden}. The same file gets it
  right twice and wrong once. And notice what happened to \Eng{guard}: it was dropped, not used. The authoritative English word was
  sitting right next to the word being guessed at.

  That pattern is everywhere. \Emph{swedish} \Eng{english} occurs at about 55 places across some twenty files, grundtyper and basic
  types, kompanjonsobjekt and companion object, aktiveringspost and stack frame, and so on. Every one is a term I have already
  translated by hand.

  So the fix I would like is structural: in the English mirror, render \Emph{X} \Eng{Y} as \Emph{Y}. One rule, it uses terminology I
  have already committed to, it fixes gard along with about fifty other terms that currently depend on the model guessing a Swedish
  compound correctly with no context, and it needs no glossary or override entries at all. It is the same choice we made on the links
  checker recently, structural rule rather than a list of known cases. Please send it as its own issue and PR rather than folding it
  in here, and take the credit for finding it.

  On line 191 and inlineCodeRe, question three. Do not widen it. You were right to hesitate. The pipe delimiter exists precisely so
  that code can contain curly braces, and there are around ninety \code|...| sites including class Klassnamn(parametrar){ medlemmar }
  in w05, package p1 { object A; package p2 { object B }} in w04, and { uttryck1; uttryck2; ... uttryckN } in w02. A brace-based
  regex cannot parse those at all.

  The blast radius also points the wrong way. Almost all of those ninety sites are Scala keywords and standard library names,
  xs.map(f), sealed, override, and they must never be renamed. Widening would expose all of them to the renamer in order to reach a
  handful.

  Line 191 should be hand-clamped instead, which matches the direction above and follows a precedent already in the repo. See
  lect-w10-override.tex:252-254, which does exactly this with pipe-delimited code:

  \ifswedish\code|enum Färg { case Röd, Svart}|\else\code|enum Colour { case Red, Black }|\fi

  Please clamp line 191 the same way, either here or in a follow-up, whichever you prefer.

bjornregnell:
Following up on my own comment above, because I understated the \Eng{} idea and the numbers change the design.

  I said the rule was \Emph{X} \Eng{Y} and that it would fix gard plus about fifty other terms. That was measured on only one of the
  shapes. Counting lines in slides/, a line with two glosses counted once:

  ┌───────────────────┬───────────────────────────────────────────┬─────────────────┐
  │       shape       │                  example                  │ lines in slides │
  ├───────────────────┼───────────────────────────────────────────┼─────────────────┤
  │ \Emph{X} \Eng{Y}  │ \Emph{gard} \Eng{guard}                   │ ~55             │
  ├───────────────────┼───────────────────────────────────────────┼─────────────────┤
  │ \Alert{X} \Eng{Y} │ \Alert{fabriksmetod} \Eng{factory method} │ 8               │
  ├───────────────────┼───────────────────────────────────────────┼─────────────────┤
  │ bare X \Eng{Y}    │ en klass är en mall \Eng{template}        │ ~50             │
  └───────────────────┴───────────────────────────────────────────┴─────────────────┘
    
  The compendium carries many more of all three, enough that a full listing ran past 32 KB. So \Emph{} alone covers roughly half the
  cases in slides and a smaller share corpus-wide. The rule needs generalising.

  The bare shape is the awkward one, because nothing marks where the Swedish term starts. Compare en klass är en mall \Eng{template},
  one word, with Repeterade parametrar \Eng{repeated parameters}, two, and principen om enhetlig access \Eng{uniform access 
  principle}, three. A rule that reaches backwards has to guess how far, and a wrong guess damages a sentence that is currently
  correct. That is worse than the defect we are fixing.

  So here is what I would like to propose instead, and I think it is cleaner than my first suggestion.

  For the bare cases, I brace the term in the source:

  en klass är en mall {template-term} \Eng{template}
  that is, {X1 X2} \Eng{Y1 Y2}. A brace group is typographically inert in running text, so nothing changes in the Swedish output, but
  the term becomes delimited exactly the way \Emph{} already delimits it.

  The payoff is that all three shapes collapse into a single rule. The term is simply the brace group immediately preceding \Eng{},
  whether or not it carries a macro name. \Emph{X}, \Alert{X} and plain {X} are then one case, not three, and the transform is:
  replace the contents of that group with the contents of \Eng{}, and drop the gloss.

  Two things I checked before proposing it. There is no occurrence anywhere in slides of \code{...} immediately followed by
  \Eng{...}, so the rule cannot rewrite verbatim code by accident. And there is no existing bare {...} \Eng{...} either, so the
  convention starts from zero with nothing legacy to disambiguate.

  For terms I have not braced yet, I would rather have a check than a guess. Assert that the English rendering of a unit containing
  \Eng{Y} actually contains Y. It transforms nothing, so it cannot corrupt anything, and it fails exactly where the pipeline
  disagreed with my own stated terminology, which is the gard case. Nobody knows how often that happens today, and I would like to.

  That also gives two numbers that move in opposite directions and are worth committing as baselines once measured: braced glosses
  may only rise, and unmatched glosses may only fall.

  Suggested order, if you want it: the gate first, since it sizes the problem and tells us which bare terms are actually worth
  bracing, then the unified brace-group transform, then I brace the remainder over time as the gate points at them. Bracing all of
  them by hand up front is work I would rather not do blind.

  Does the unified brace-group rule look right to you, or do you see a case it breaks?

hmiddelk:
Thanks — and thank you for measuring the ratchet yourself rather than making me do it.

## vattnig: fixed and verified

Pushed. The `codeStr` key is now `"vattnig" -> "watery"`, rebased onto 366e2d0e so it is keyed against your corrected source. Verified where it actually matters — the generated mirror, not the gate:

```
lect-w11-context-en.tex:433  object Cucumber extends Vegetable(42), HasTaste("watery")
lect-w11-context-en.tex:434  object Tomato   extends Vegetable(43), HasTaste("tart")
```

with `fallbacks: 9` / `OK: 9 within 9` and `[overrides] OK: 45 within 45` on that base.

Worth reporting how I nearly shipped it broken, because it argues your point better than agreeing with you does. I first changed the key and ran the full gate **without** having actually picked up your commit — the rebase had not landed. Every gate passed, 9 within 9 and 45 within 45, and the mirror still said `HasTaste("vattning")`: the old source spelling still matched the old key, and my new key matched nothing at all. Both states are green and only one is correct. The only evidence that distinguished them was reading line 433.

## Yes to a CodeGlossary orphan check

Please open it. `codeStr` and `id` keys have no check at all today, and between us we have now produced three dead ones in a week: `"vattning"` after your fix, my two `Overrides` fragment keys that matched nothing (`"fel!"`, `"kontravariant med -"`), and the `MinKontext` entry which becomes dead the moment you correct that slide. The Overrides ratchet caught none of them, because it does not look at CodeGlossary.

One design note if you do open it: the orphan notion differs per map. An `Overrides` key is either matched or not, but a `perFileId` key legitimately matches nothing in a full run if its file has no Scala-code env left to rewrite, and `id` keys are global by design, so many are unused in any given corpus. The signal worth gating is narrower: a key whose *Swedish spelling does not occur anywhere in the source it is scoped to*. That catches `"vattning"` after your rename, and would not have flagged `"hej"`.

## On #981 — your revert answers most of my review

I had drafted a warning that bracing would change the unit text and therefore the cache key, and you had already demonstrated it in 366e2d0e with four concrete regressions. So I will only add what is left.

**The already-delimited cases can be fixed with no source edit and no cache churn at all.** Your probe says 138 of 307 glosses are macro-delimited and 40 of the 108 defects sit in that set. For those the source needs no change, so the risk you hit does not arise — provided the transform runs *after* the cache lookup rather than before it. Concretely: key the cache on the untouched source unit exactly as today, take the translated result, and then substitute into it using the source as the guide — the source unit says the *n*th `\Emph` group is glossed `guard`, so replace the contents of the *n*th `\Emph` group in the translated unit and drop any gloss remnant. If that correspondence cannot be established, because the number of groups differs or a macro was dropped, change nothing and let the probe report it. Fail-closed means the worst case is "no improvement", never your `"When you {shadow}"`.

**And the bare cases may be cheaper than you think.** Bracing cost you a cache miss because the key is the raw unit text. If the key were normalised to strip a *bare* brace group immediately preceding `\Eng{` — only that shape, not `\Emph{}`, and not the `\fi{}` space guards apply-b0d emits — then `{template-term} \Eng{template}` and `template-term \Eng{template}` would key identically, bracing would become genuinely free, and 888d0af6 would have been the no-op you expected. I have not checked where keying happens, so treat that as a question rather than a claim: is the key the raw unit text, and is there a normalisation point where this would be one narrow rule?

Two smaller things on the probe itself. Its correspondence is by line number and it already reports 7 drifted lines; a transform cannot use line numbers, so it will need unit-level correspondence anyway, which is the same machinery as the `\Emph`-ordinal idea above. And treating 108 as an upper bound is right — `traversera` to `traversing` is fine, and a strict containment test will also flag `\Eng{basic types}` against a correct "basic type", so whatever number you ratchet on probably wants singular/plural tolerance or it will freeze at a value that is mostly noise.

Since the probe exists, the split I would suggest is: you keep the measurement half, and I take the transform for the already-delimited subset, fail-closed, with your probe as its acceptance test — the 40 delimited defects should fall and nothing else should move. Say if you would rather own that too.

## Line 191 and direction

Line 191: agreed, and I will hand-clamp it as its own follow-up using your `lect-w10-override.tex:252-254` precedent rather than folding it in here. Thank you for the `\code|...|` numbers — around ninety sites, keywords and standard-library names among them, is a much better reason not to widen `inlineCodeRe` than my hesitation was.

Direction: understood, and this is the last glossary-driven batch.

That said, reading the built English compendium turned up a gap the direction affects. w03 has no `\ifswedish` clamps at all, so `lect-w03-functions` and `w03-functions-exercise` render Swedish identifiers in the English build: 3.1.6 declares `förnamn`/`efternamn`/`förnamnFörst`/`ledtext`, 3.1.12 has `skrivLedtrådOmEjRätt`/`gissning`/`hemlis`, and the chapter-3 task REPLs have `öka`, `skrivNamn`, `hälsa`, `fleraAnrop`. 3.1.35 is the one worth looking at yourself: it is mixed rather than Swedish, with the slide title and the interpolated strings in English while the identifiers stayed Swedish, and `är för stort` / `är för litet` still in the strings.

Under the new direction that is `\ifswedish` work rather than a glossary cluster. Shall I open an issue for it so it is at least recorded, and leave the clamping to whenever it suits you?

bjornregnell:
Thank you, this is careful work and the four-mechanism table is the part I want to keep: that picking the wrong mechanism is a silent no-op is exactly the failure class worth naming, and it is why the orphan gate earns its keep.

  All entries ratified as proposed, including both judgement calls:

  - framkalla → elicit is right, and your reason is the correct one: clamping to summon would make the next slide's sentence
    tautological.
  - tulipanaros left Swedish is right. It is the joke and the link target, not an example name. Leave it.
  - The three global promotions are fine given you verified the tokens occur nowhere else, and token-exact matching keeps Ros off
    Rosenqvist.
  - The MinKontext mirror-not-fix is the right call, and yes please, send the #968 PR that fixes the Swedish and drops that entry in    the same change.
   
  Verified after merging, on master: [overrides] OK 45 within 45, [cache-only] OK 9 within 9, defined override keys 395 → 399 with  the orphan count unchanged at 45. So all four of your keys match live units and the merge added no orphan debt, exactly as your own
  run reported.

  On your direction question, and I have a new reason to answer it the way I now do.

  I said on #969 that I leaned toward clamped \ifswedish for future Swedish-side changes. I am revising that, and the glossary-driven  style you used here should be the default for renames.

  The reason is something measured today rather than reasoned. I ran a small consistency sweep over the keyboard key names in the  Swedish source, which is exactly the shape of change \ifswedish clamping involves: touch the Swedish, regenerate. Changing a unit's
  Swedish invalidates its cache row, the model is called again, and two of the three re-translated units came back damaged. One came  back as verbatim Swedish.

  Both gates stayed green through that. Fallbacks held at 9 and orphans at 45, because a model result that happens to be Swedish is  cached as a success, not counted as a fallback. So [cache-only], whose job is to prove the English side resolves, is structurally   blind to it. The only thing that caught it was the swedish-left figure pdfCompendiumEn prints, which rose from 1187 lines to 1191.   I repaired it by pinning the two units as reviewed overrides, and the count came back to exactly 1187.

  That cuts directly at the trade-off you framed. Your glossary approach does not touch the Swedish source, so it invalidates no   cache rows and cannot trigger that failure mode at all. \ifswedish clamping does, on every use.

  So: glossary-driven for renaming identifiers, strings and comments; \ifswedish reserved for where the two languages genuinely need    different content, which is what it is actually for. I will maintain the entries. The per-name ratification cost is real but it is
  cheaper than a silent Swedish leak that no gate reports.
 
  Two consequences worth stating, since they are mine to own rather than yours:
 
  1. The gates prove the mirror resolves, not that it is English. Worth remembering before treating a green mirror gate as proof.
  2. I think a model result identical to its Swedish input should not be cached as a success. That is the hole. The comparison is      already in hand at cache-write time, so it looks cheap. I will file it.
    
  The \code|...| regex widening and the w03 gap you flagged both stay open and are not yours to carry.

  Assistance note: I used Claude Code to verify the post-merge gate numbers and to run the sweep whose measurement is described  above. The direction decision is mine.

PR #978, opened 2026-09-18
Fixed #977

### LINKED ISSUE #977: Workspace errors when folder opened in vscode
If you currently open the workspace folder in vscode, you get the following errors:
1. Overlap in def names for run in most wxx_labname folders. 
Recommended fix: Create a workspace.code-workspace for vscode that separates them into different projects.
2. w04_blockmole missing introprog (vscode cannot find introprog even if it's there). 
Recommended fix: Add a project.scala file to it same as w06 (the continuation).
3. Mismatch in `//> using scala 3.x` between 3.5 and 3.8.3. This causes metals to download older scala which is not wanted.
Recommended fix: just keep it to 3, it will use the current installed version which should work.
4. in w08_latex: NewtonRalph uses same def name as another file in same directory tree.
Recommended fix: rename it to something else (like `exec`)

Annie-Wb:
This implements the exact fixes recommended by issue #977 
So it fixes #977.


Annie-Wb:
Introprog hints that it is outdated:
```bash
[hint] ./project.scala:3:15
[hint] "introprog is outdated, update to 1.5.0"
[hint]      introprog 1.4.0 -> se.lth.cs::introprog::1.5.0
[hint] //> using dep se.lth.cs::introprog::1.4.0
[hint]             
```

I was going to add that change too before merge.

bjornregnell:
aha; another PR is welcome :)

Issue #983, opened 2026-09-18
autotranslate: some English mirror lines are not translated at all, and nothing reports it


bjornregnell:
Some lines in the English mirror are not translated at all. They come through as Swedish prose, or half Swedish and half English on the same line, and nothing reports it.

This is a different defect from #981. That one is about the pipeline choosing a worse English term than the one the source supplies. This one is about lines where no English was produced.

## Confirmed examples

Each is a line in the Swedish source carrying an `\Eng{}` gloss, paired with the same line number in the generated mirror.

**Half translated, one clause each way** — `slides/body/lect-w11-context.tex:66-67` against `slides-en/body/lect-w11-context-en.tex`:

```
  \item med \code{T <: U} blir \code{U} en övre gräns  för \code{T}\\\code{U} is the 'highest' possible type t...
  \item med \code{T >: L} blir \code{L} en undre gräns  för \code{T}\\\code{U} is the 'lowest' possible type t...
```

The first clause stayed Swedish, the second became English, in one line. The `\Eng{upper bound}` and `\Eng{lower bound}` glosses on those lines were dropped rather than used.

**Whole line untranslated** — the mirror repeats the Swedish verbatim:

| source | mirror line reads |
| --- | --- |
| `slides/body/lect-w05-equality.tex:12` | `\item \Emph{Innehållslikhet}, ä.k. strukturlikhet : två referenser anses lika...` |
| `slides/body/lect-wjava-body.tex:1656` | same line, same leak (the two files share this content) |
| `slides/body/lect-w01-intro.tex:442` | `\item Scala och Java är s.k. ... \Emph{statiskt typade} språk, vil...` |
| `slides/body/lect-wjava-body.tex:1524` | `\item Speciella regler för implicita konverteringar, s.k. ''auto-boxing''` |

`lect-w01-intro.tex:442` carries two glosses, `\Eng{compile time}` and `\Eng{run time}`, and neither reached the mirror.

**`\Eng{}` itself survives into the mirror.** `slides-en/` contains **12** occurrences of `\Eng`. Some of those are inside `\ifswedish ... \else` Swedish branches and belong there, but the presence of any is worth knowing, because it means "the mirror contains no `\Eng`" is not true today and a check on that is not measuring nothing.

## How this was found

[...]

Issue #981, opened 2026-09-18
autotranslate: use the source's own \Eng{} gloss as the English term, via one brace-group rule


bjornregnell:
The English mirror mistranslates a term that the Swedish source already translates by hand, a few characters to the right of the word it got wrong.

## The symptom

`slides/body/lect-w06-matching.tex:247` reads:

```latex
Man kan stoppa in en s.k \Emph{gard} \Eng{guard} innan pilen \code{=>} för att villkora matchningen
```

The generated `slides-en/body/lect-w06-matching-en.tex` renders:

```
246: \begin{Slide}{Matching with guard}            <- correct
247: One can insert a so-called \Emph{garden}...   <- wrong
267: \code{case}-branch with guard gives...        <- correct
```

The same file is right twice and wrong once. It fails exactly where the term sits alone inside `\Emph{}`, which is the isolated child-unit case: translated with no surrounding sentence, `gard` reads as a truncated English word or as Swedish `gård`, and the model guesses.

Second occurrence, same shape: `slides/body/lect-w07-seq.tex:350`. There are none in `compendium/`.

## The point

`\Eng{guard}` was **dropped, not used**. The authoritative English term is present in the source, written by me, and the pipeline discards it and then guesses it wrong.

This is not a two-instance problem. `\Emph{<swedish>} \Eng{<english>}` is a pervasive convention: roughly **55 occurrences across about 20 files** in `slides/` alone. A sample:

| Swedish | the source's own `\Eng{}` |
| --- | --- |
| `grundtyper` | basic types |
| `kompanjonsobjekt` | companion object |
| `aktiveringspost` | stack frame |
| `typkonstruktor` | type constructor |
| `inkapsling` | encapsulation |
| `överlagring` | overloading |
| `platshållare` | placeholder |
| `egentyp` | self type |

Every one of these is a term I have already decided. Today each depends on the model translating a Swedish compound correctly, out of context, on every run. `gard` is simply the one that has been noticed.

## First cut, and why it is not enough

The obvious rule is to render `\Emph{X} \Eng{Y}` as `\Emph{Y}`, taking `Y` verbatim and dropping the gloss, which is redundant once the surrounding text is English. (The mirror already drops `\Eng{...}`; it just does not use it first.)

That covers only about half the corpus. The counts below generalise it, and the generalised matcher is the one to implement.

## `\Eng{}` is often NOT preceded by `\Emph{}`

Measured in `slides/` (line counts, so a line carrying two glosses counts once):

[...]

PR #984, opened 2026-09-18
autotranslate/scratch: probe whether the mirror uses the source's own \Eng{} gloss (measurement half of #981)

bjornregnell:
The measurement half of #981, as a scratch probe beside `unit-probe.scala` and `bookmark-probe.scala`.

It answers a question none of us can currently answer: **how often does the English mirror contradict terminology the Swedish source has already committed to in an `\Eng{}` gloss?** It transforms nothing, so it cannot corrupt output, and it needs no source edits to run.

## Measured on this branch

```
\Eng glosses found: 307   (delimited 138, bare 169)

  ok              158   (delimited  84, bare  74)
  TERM-MISSING    108   (delimited  40, bare  68)
  SUSPECT-SV        9   (delimited   0, bare   9)
  UNTRANSLATED      2   (delimited   1, bare   1)
  UNALIGNED         7   (delimited   2, bare   5)
  EMPTY-MIRROR     23   (delimited  11, bare  12)
```

**The number that matters for sequencing: 40 of the 108 are ALREADY delimited by a macro.** Those need no source edit at all, so fixing them costs no cache invalidation and no model run. That is the cheapest and largest first step for #981, and it is why I would not start with bracing.

## It reports an upper bound, deliberately

`TERM-MISSING` also catches acceptable rewordings — `traversera \Eng{traverse}` rendering as "traversing" — and inflection differences. On a hand-checked slice of 20 rows, roughly two thirds were real defects. The probe says so in its own output rather than letting the number be quoted as a defect count.

## Four things it does that a naive comparison gets wrong

Each of these was learned by getting it wrong first; an earlier version of this probe reported all four as defects.

1. **Skips `%`-commented source lines.** An untranslated comment is never typeset. This alone removed 21 spurious rows.
2. **Strips LaTeX before comparing.** `\Emph {partially applied} functions` does contain the gloss "partially applied functions" — a substring test on the raw line does not see it and calls a correct translation broken.
3. **Separates a byte-identical mirror line from a merely Swedish-looking one.** A line whose English prose is perfectly good can still carry `\code{Färg}` where an identifier is deliberately unclamped. Counting diacritics alone over-reported untranslated lines by roughly 15×: 30 became 2 hard leaks plus 9 weaker suspects.
4. **Separates line-number drift from translation failure.** Source and mirror line numbers align for most files but not all; rows landing on a bare `\begin{enumerate}` are an alignment artefact, not a defect.

## Bearing on #983

The two hard leaks it finds are real and worth looking at:

```
slides/body/lect-w01-intro.tex:451      \Eng{basic types}
slides/body/lect-w02-codestruct.tex:157 \Eng{vector}
```

And it refines #983's headline figure, which I filed as "30, and an upper bound I have not reduced". This reduces it: **2 byte-identical leaks and 9 lines carrying Swedish outside code regions.** The partially-translated cases named in #983, such as `lect-w11-context.tex:66-67` where one clause is Swedish and the next is English, are still real — they land in `SUSPECT-SV` rather than `UNTRANSLATED` because the line is partly translated rather than untouched.

## Not done here

No gate, no baseline, no CI wiring. This is the instrument; whether it becomes a gate, and with which of its six verdicts as the gated number, is a decision for #981 and #983 rather than something to slip in with a probe. If it does become a gate, the baseline should be measured and committed rather than guessed, and proved able to fail first — the same discipline as `override-orphans-baseline.txt` and the heading-map floor.

Scala 3.9.0 rather than the 3.3.4 the sibling probes pin, since 3.9.0 is the new LTS. Worth bringing those two along at some point; not in this PR.

## Assistance note

Written with Claude Code, which built the probe, ran it, and wrote this description. The four corrections above came from its earlier version producing wrong numbers that I checked by hand against the files. I have run it on this branch and read the rows it flags.


bjornregnell:
Two one-line defects in the probe, found while reviewing it, both worth fixing before its numbers get quoted as a sequencing   decision for #981.
  
  1. A dead condition. In def unaligned, m.startsWith("\\item]") can never fire: the LaTeX token is \item[], not \item]. So UNALIGNED      under-reports, and those rows fall through to SUSPECT-SV or TERM-MISSING instead.
  2. A bias in the headline split. In def glossesIn, val pre = line.substring(0, k).stripSuffix(" ") strips exactly one trailing      space, and pre.endsWith("}") then decides whether a gloss counts as delimited. Two spaces, or a tab, before \Eng{ therefore      misclassifies an already-delimited gloss as bare. That skews the delimited-versus-bare ratio, which is the number used to argue      that the already-delimited glosses are the cheapest and largest first step. stripTrailing() fixes it.
    
  Lower severity, but worth noting before this ever becomes a gate: the TSV writer does not escape tabs or newlines in the eng,   before or mirror-excerpt fields, so a tabbed source line corrupts the columns; swedishInProse strips \code{...} with a non-nesting   [^}]*; and flatten leaves non-letter control sequences such as \& and \\ in place.

  Nothing consumes the probe, so landing it and fixing in the follow-up that turns it into a gate is defensible. Defect 2 should be   fixed before the delimited-versus-bare figures are quoted as a decision, though.

PR #985, opened 2026-09-19
Fix typo in dod-lab-linux

antonmunsin:
Fixed two typos in the linux dod lab where students often get stuck:
- "Linux" to "Unix" in the part where grep was used on the file "example.txt" ("Linux" appear 0 times)
- Fixed whitespaces that does not compile correctly in code blocks, in the same part of the lab


bjornregnell:
@mnordahl Ok to merege?

bjornregnell:
Good! many thanks @antonmunsin

PR #986, opened 2026-09-20
lect-w11-context: clamp two Swedish-showing spans, fix the L bound, stop the Kovarians frame dropping its tail

hmiddelk:
Three small source changes to `slides/body/lect-w11-context.tex`, all model-free, one commit each so any of them can be dropped independently. Two of them are things you asked for in #976; the third fell out of verifying them.

## 1. Clamp the two places the English side showed Swedish

**The type-bound items (66-67).** The source line is Swedish and names its own English term, as `\Emph{gard}`-style `\Eng{upper bound}`. The mirror split the item at the `\\` and translated only the second half, so an English reader got the Swedish clause followed by an English paraphrase of it:

```
med T <: U blir U en övre gräns för T
U is the ’highest’ possible type that T can be (compare ’less than or equal to’)
```

The first half cannot improve on its own. `translate-cache.tsv` holds an identity row keying `med __C0__ blir __C1__ en övre gräns  för` to itself, so it is a cache **hit** and renders Swedish with or without a model backend — one of the stale key==value rows from #960. Clamping sidesteps both that and the discarded gloss (#981).

**The covariant-enum slide (191).** As you asked. Clamped in the style of `lect-w10-override.tex:252-254`, with English names matching the glossary entries in #976 (`Toilet`, `Occupied`, `Vacant`), so both branches of the same demo agree once that merges.

## 2. The lower bound is L, not U

In the type-bound list the second clause of the `>:` item said that `U` is the lowest possible type `T` can be. For `T >: L` it is `L`; `U` is the upper-bound example on the line above. The slip is in the Swedish and the mirror reproduced it faithfully.

Fixed on **both** sides, deliberately: fixing only the English would recreate exactly the inconsistency you objected to over `vattning`, where the Swedish kept an error the English had quietly corrected. Separate commit, so drop it if you would rather make the Swedish change yourself.

## 3. Shrink the Kovarians frame so its tail is not dropped

This is the one that was not on the list, and it is why clamping 191 was not observable.

The English deck was losing the end of that frame: page 14 stopped mid-item at `Ledig is a Toalett[Nothing]`, missing that item's own trailing `here:` **and** the entire enum item after it. The Swedish deck renders both on the same page. So clamping 191 changed a span the English deck was not printing at all.

Ruled out by measurement, not reasoning:

- **Not my clamp.** With line 191 restored to its upstream form and the frame untouched, the regenerated mirror carried the unclamped enum line and the English deck still showed neither it nor `here:`.
- **Not the mirror.** The generated `-en` file contains the item; only the PDF lacks it.

The cause is the frame overflowing, in English only, because English prose is longer: that frame packs an itemize at `\SlideFontSmall` with a `Code` block, a `REPL` block and three `\pause` steps, and unlike its neighbours at 39, 59, 130 and 311 it has no font reduction. The English deck reported 41 overfull `\vbox` warnings against the Swedish deck's 27.

`[shrink]` needs no new machinery, since `Slide` passes its optional argument straight to `frame` (`slides/lecturenotes.cls:147`), and beamer only scales a frame that actually overflows.

## Verification

Full `autotranslateProject/run --cache-only` on this base: `fallbacks: 9` / `OK: 9 within 9`, `[overrides] OK: measured 45 within baseline 45` — the clamps move neither number, which is the point: a clamped span is masked rather than translated, so it adds no cache demand.

Built deck, `slides-en/lect-w11-en.pdf`:

```
with T <: U, U becomes an upper bound for T
U is the ”highest” possible type that T can be (compare ”less than or equal to”)
with T >: L, L becomes a lower bound for T
L is the ”lowest” possible type that T can be (compare ”greater than or equal to”)
...
here:
enum Toilet[+T] { case Occupied(x: T); case Vacant }
```

with zero Swedish-branch leakage, and the overfull count down 41 → 37. The Swedish deck is unchanged at 27 overfull before and after, confirming `shrink` is inert where content already fits.

One expected artefact while #976 is unmerged: `Ledig is a Toalett[Nothing]` on that same line is prose `\code{}` renamed by the #976 glossary, which is not on this branch. It resolves when that merges.

## Not fixed here

The remaining 37 overfull frames in this one deck. If any of them drop content the way this one did, the English decks are lossy and nobody has measured how much — filed as its own issue, since 37 hand-placed shrinks across 19 decks is the wrong shape of answer and an overfull warning is not by itself the signal (the Swedish deck's 27 apparently cost nothing).

## Assistant note

*(per the "Making contributions" guidelines in the README)* I used Claude Code throughout: it found the identity cache row behind the type-bound item, proposed the clamps, and ran the control build that falsified its own first theory about the dropped content — that the clamp was to blame. I checked the results in the built PDFs rather than the diff, and every number above comes from a build log or a `pdftotext` extract.


hmiddelk:
`English-mirror-gate` is red on this PR, and it is inherited from master rather than caused by these three commits. Recording the numbers so the tick is readable.

This PR's merge build:

```
done. model calls: 0, fallbacks: 25, overrides: 485, cache: 15513
[overrides] OK: measured 45 within baseline 45
[cache-only] FAIL: measured 25 exceeds committed baseline 9
```

Master alone, at c290d85e:

```
done. model calls: 0, fallbacks: 25, overrides: 485, cache: 15513
[cache-only] FAIL: measured 25 exceeds committed baseline 9
```

Identical, so this branch adds **zero** fallbacks. Master has been red at both of today's commits, b2cfec83 and c290d85e, and the growth looks like the familiar shape: the cache gained 16 rows over our base (15497 → 15513) while the count rose by exactly 16 (9 → 25), with `slides/body/lect-w04-objects.tex` among the changed files. Same situation as #969 — new prose needing a model pass, which only you can run.

Measured locally on this branch before master moved, base 366e2d0e:

- Full `autotranslateProject/run --cache-only`: `fallbacks: 9` / `OK: 9 within 9`, `[overrides] OK: 45 within 45`.
- After the `[shrink]` commit, a scoped `--only w11` run: `fallbacks: 0`.

That second one matters for reviewing commit 3, since `[shrink]` edits the `\begin{Slide}` line and it would be reasonable to worry that an optional argument breaks title extraction and leaves the title untranslated. It does not: w11 measures 0 fallbacks with it in place, and the deck renders the English title.

Worth stating why clamping cannot move these numbers in either direction: a clamped span is masked rather than translated, so it adds no cache demand and removes none — which is also why the identity row behind the type-bound item stays in the cache, harmlessly unused, rather than needing a purge.

No action wanted from you on this; the PR is reviewable as it stands, and its own gate will go green whenever master's does.

bjornregnell:
I have updated this branch against master and CI is now green on all three jobs (run 35902528214). Please read the previous red as  stale rather than as anything you did.

  Why it was red: master itself was failing from 2026-09-20 until today, on two stacked ratchet breaches: the cache-only count had   drifted to 31 against a ceiling of 9 after some Swedish edits of mine, and an override key orphaned by #985 pushed the orphan count   to 46 against 45. Because the orphan gate throws first, the cache-only number was hidden. Every branch inherited that red, and  your branch's mirror-gate output was byte-identical to its base's, which is how I know it added nothing of its own.
 
  Both are fixed on master now, and #976 is merged.

  Three review points, none blocking:

  1. slides/body/lect-w11-context.tex:66-67 — the English branch uses the Swedish ''…'' for both quotes, so it renders as ”highest”.     English convention is ``highest''. One character, worth doing while the line is open.
  2. :161, the [shrink] — the mechanism is sound, Slide passes its optional argument straight to frame. My hesitation is that this     touches the Swedish source of truth, and "Swedish overfull count unchanged at 27" is a count of frames, which does not establish
     that this frame was not already among the 27 and is now silently scaled. Could you check that one page before and after? This     is the one I would like answered before merging.
  3. A dead cache row. Clamping 66-67 orphans the translate-cache.tsv identity row you diagnosed yourself. Nothing gates cache rows,     so it will sit there. Either prune it here or let us track it.

  Also, a figure to correct for the record rather than a problem: the body reports fallbacks: 9 on this base, while CI measured 25.
  That is explained by master's drift landing between your local run and CI, but the number as written would mislead whoever quotes  it next.
  
  One thing to watch when this merges, learned the hard way today: your clamps change Swedish source text, which invalidates those  units' cache rows and causes re-translation. I hit a case today where the model returned a unit unchanged in Swedish, which is  cached as a success and is therefore counted as neither a fallback nor an orphan, so both gates stayed green while the English  mirror silently got worse. The only signal was the swedish-left figure from pdfCompendiumEn, which is printed but gated by nothing.
  Written up as a comment on #983.
  
  So after this lands, please do not stop at green gates. Compare swedish-left against the current baseline of 1149/26139 lines, 4.4%   (it was 1187 before your #976 merged, which removed 38 Swedish lines from the compendium, so thank you for that).

  Assistance note: I used Claude Code for the master-side diagnosis and repair described above, and for checking this branch's inherited status. The review points are mine.

bjornregnell:
I have answered my own [shrink] question by building it rather than asking you to.
 
  Swedish lect-w11, Overfull \vbox count in the pdflatex log: 27 without [shrink], 27 with it. So [shrink] is inert on the Swedish   side. Beamer only scales a frame that overshoots, this one does not, and the Swedish page is unchanged. Your "27 unchanged" figure   is also confirmed as the right number for that deck.

  My reservation was that a frame count cannot establish that this particular frame is unaffected, only that the total did not move.   That was worth checking and it checks out, so the reservation is withdrawn. Merging.
 
  Two small things I would still take, either here or as a follow-up, neither blocking:

  1. :66-67 — the English branch uses the Swedish ''…'' for both quotes, rendering as ”highest”. English convention is ``highest''.
  2. The dead cache row that clamping 66-67 orphans. Nothing gates cache rows, so it will sit there quietly. Prune here or let us      track it.
     
  And for the record, corrected: the fallbacks: 9 in the body versus CI's 25 was master's drift landing between your run and CI's,   not anything of yours. Master is green now, and the current swedish-left baseline after your #976 merged is 1149/26139 lines, 4.4%,   improved by 38 lines by that merge.

